PR #20, opened 2026-04-24
Fix TypeScript validation blockers

SweetSophia:
## Summary
- Generate Prisma client during diagnosis to confirm generated types were the source of Prisma export errors
- Fix the remaining TypeScript error in the obsidian sync test by importing `readdirSync`
- Replace lint-blocking `require()` usage with static imports and make the API key permission variable const

## Validation
- `npx tsc --noEmit`
- `npm run lint` passes with existing warnings only

## Summary by Sourcery

Resolve remaining TypeScript and linting blockers by cleaning up test utilities and script typing.

Bug Fixes:
- Eliminate CommonJS require usage in Obsidian sync tests to satisfy TypeScript and linting rules.
- Tighten the API key creation script by making the permission argument immutable to align with expected typing and usage.

### DIFF
--- a/scripts/create-api-key.ts
+++ b/scripts/create-api-key.ts
@@ -9,7 +9,7 @@ import { closePrisma, prisma } from "./_prisma";
 
 async function main() {
   const name = process.argv[2];
-  let permissions = (process.argv[3] || "WRITE") as "READ" | "WRITE" | "ADMIN";
+  const permissions = (process.argv[3] || "WRITE") as "READ" | "WRITE" | "ADMIN";
 
   if (!name) {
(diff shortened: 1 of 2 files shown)


SweetSophia:
Correction to the previous comment (shell stripped inline code formatting):

Addressed the valid review feedback in fda68c0:

- Updated the test helper safePath to mirror production boundary checks using a normalized vault path.
- Replaced fragile temp-directory cleanup with recursive rmSync(..., { force: true }).
- Switched js-yaml to a default import for consistency.
- Changed frontmatter serialization to use an ordered record via FM_KEYS, matching production behavior and removing the unused ordered warning.

Validation:
- npx tsc --noEmit
- npm run lint (passes with existing warnings only)
- npx tsx src/__tests__/obsidian-sync/index.test.ts

PR #24, opened 2026-04-25
Add Noosphere memory provider adapter

### LINKED ISSUE #9: Build Noosphere provider adapter
Part of #5\n\nImplement Noosphere self-query support through the same provider abstraction used for external memory systems.\n\n## Acceptance Criteria\n- Noosphere can be queried through the universal provider layer\n- curated knowledge maps cleanly into normalized results

SweetSophia:
## Summary
- Add a Noosphere memory provider implementing the universal `MemoryProvider` contract.
- Map wiki articles, topics, tags, lifecycle status, confidence, and provenance into normalized `MemoryResult` objects.
- Export the Noosphere provider class, factory, and metadata option types from the memory module.

## Verification
- `npm run lint` (passes with existing warnings)
- `npx tsc --noEmit`

Closes #9

## Summary by Sourcery

Add a Noosphere-backed memory provider integrating wiki articles into the universal memory system.

New Features:
- Introduce a Noosphere memory provider implementation exposing search, fetch-by-id, and scoring capabilities over wiki articles.

Enhancements:
- Export the Noosphere memory provider factory, class, and related metadata option types from the memory module for external use.

### DIFF
--- a/src/__tests__/memory/noosphere-provider.test.ts
+++ b/src/__tests__/memory/noosphere-provider.test.ts
@@ -4,30 +4,37 @@
  * Run with: npx tsx src/__tests__/memory/noosphere-provider.test.ts
  *
  * Tests cover:
- * 1. Pure function unit tests (mapConfidenceScore, mapCurationLevel, mapRecencyScore, normalizeRank, normalizeOffset)
- * 2. Constructor validation
- * 3. search() with mock Prisma (empty query, disabled config, successful search)
- * 4. getById() with mock Prisma (found, not found, disabled, canonical ref stripping)
- * 5. score() edge cases (non-noosphere results, missing fields, aggregate calculation)
- * 6. resolveSearchLimit priority chain
+ * 1. Constructor validation (descriptor, config overrides, factory)
+ * 2. search() edge cases (empty query, disabled config, no results)
+ * 3. getById() (found, not found, disabled, canonical ref stripping)
+ * 4. score() (non-noosphere guard, aggregate, recency decay)
+ * 5. Confidence and curation level mapping via getById
+ * 6. Descriptor metadata
  */
 
-// ─── Inline access to private-scoped helpers via re-export ──────────────────
-// The helper functions are module-scoped, so we test them through the provider's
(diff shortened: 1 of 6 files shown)


SweetSophia:
Addressed the valid review feedback in 63eee0d:

- Guarded rank normalization against non-finite values.
- Switched recency scoring to true half-life decay.
- Stopped defaulting missing relevance to 1 in provider-local scoring.
- Honored disabled provider config in getById().
- Removed the redundant post-query sort while preserving SQL result order.
- Allowed explicit maxResults overrides of 0/undefined to remove the provider-local SQL LIMIT; descriptor default remains capped at 10.

Verification:
- npm run lint (existing warnings only)
- npx tsc --noEmit

SweetSophia:
Also addressed the shared-search maintainability comment in a follow-up commit by exporting/reusing the wiki article search filter builder from the Noosphere provider. Re-ran:

- npm run lint (existing warnings only)
- npx tsc --noEmit

SweetSophia:
Final maintainability follow-up: moved the duplicated metadata cleanup helper into the shared memory type utilities and reused it from both Hindsight and Noosphere adapters.

Verification repeated:
- npm run lint (existing warnings only)
- npx tsc --noEmit

SweetSophia:
I did a deeper self-review plus an independent architecture/code review pass. No blockers found. I pushed one hardening commit (f583925) for the edge cases found:

- Avoid unbounded queries for invalid explicit limits while still allowing an explicit undefined maxResults override to remove the provider-local cap.
- Use a reduce-based max rank calculation instead of spreading rows into Math.max.
- Rename metadata sourceType to articleSourceType to avoid confusion with top-level MemoryResult.sourceType.
- Only include the aggregate-score diagnostic reason when an aggregate score is actually computed.

Verification:
- npm run lint (existing warnings only)
- npx tsc --noEmit

Remaining non-blocking debt: the full-text search CTE is still duplicated between wiki search and the provider. Filters are shared now, but extracting the whole ranked-search query would be a larger cleanup best done separately.

PR #39, opened 2026-04-29
feat(memory): add status API endpoint

SweetSophia:
## Summary
- add authenticated GET /api/memory/status
- expose safe memory provider metadata and normalized public recall settings
- add DB-free status helper tests and include them in test:memory

## Verification
- node --import tsx --test src/__tests__/memory/api-status.test.ts
- npm run test:memory
- npx tsc --noEmit
- git diff --check

## Notes
- status metadata is descriptor-based so this endpoint can be tested without DATABASE_URL and without instantiating Prisma-backed providers.

## Summary by Sourcery

Add an authenticated memory status API endpoint that exposes provider metadata and recall settings, along with supporting helpers and tests.

New Features:
- Expose a GET /api/memory/status endpoint that returns a normalized snapshot of memory provider status and recall settings.

Enhancements:
- Introduce a descriptor-based memory status snapshot helper to derive provider and settings metadata without requiring database-backed providers.

Build:
- Extend the test:memory npm script to include the new memory status API tests.

Tests:
- Add DB-free tests for the memory status snapshot helper and API endpoint.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "build": "prisma generate && next build",
     "start": "next start",
     "test": "npm run test:memory",
-    "test:memory": "node --import tsx --test src/__tests__/memory/budget.test.ts src/__tests__/memory/conflict.test.ts src/__tests__/memory/dedup.test.ts src/__tests__/memory/hindsight-provider.test.ts src/__tests__/memory/promotion.test.ts src/__tests__/memory/backfill.test.ts src/__tests__/memory/recall-orchestrator.test.ts src/__tests__/memory/settings.test.ts src/__tests__/memory/scheduler.test.ts",
+    "test:memory": "node --import tsx --test src/__tests__/memory/api-status.test.ts src/__tests__/memory/budget.test.ts src/__tests__/memory/conflict.test.ts src/__tests__/memory/dedup.test.ts src/__tests__/memory/hindsight-provider.test.ts src/__tests__/memory/promotion.test.ts src/__tests__/memory/backfill.test.ts src/__tests__/memory/recall-orchestrator.test.ts src/__tests__/memory/settings.test.ts src/__tests__/memory/scheduler.test.ts",
     "test:scheduler": "node --import tsx --test src/__tests__/memory/scheduler.test.ts",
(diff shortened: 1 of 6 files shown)


SweetSophia:
Addressed the valid review feedback in the latest commit:

- moved Noosphere provider identity/defaults/capabilities into a DB-free shared descriptor to avoid drift
- reused that descriptor from `NoosphereProvider` and the status API
- changed status auth to `requirePermission(request, ["WRITE"])`, which keeps API-key/session consistency and blocks READ-only keys
- logged the full error object on the server while keeping the client response generic
- simplified `normalizeRecallSettings(options.settings)`
- computed `allowAutoRecall` through `getEffectiveAutoRecall()` so hard provider capabilities win over config flags
- added `summaryFirst` to public settings
- fixed the duplicate `apikey` assertion and tightened secret-like field checks
- added a regression test for provider auto-recall capability gates

Verification:
- `node --import tsx --test src/__tests__/memory/api-status.test.ts`
- `npm run test:memory` (83/83)
- `npx tsc --noEmit`
- `git diff --check`

Notes:
- I did not implement dynamic production provider discovery in this PR. The endpoint remains descriptor-based by design so `/api/memory/status` stays DB-free and safe for this first slice. Dynamic provider wiring belongs with the recall API/provider registry follow-up.
- I kept `priorityWeight`, `maxResults`, and capabilities visible but raised auth from READ to WRITE+, since this is operational metadata for plugin/operator diagnostics.

SweetSophia:
Follow-up on remaining visible review threads:

- `allowAutoRecall` capability gate is fixed in `9fe41b2` via `getEffectiveAutoRecall(descriptor.capabilities, config)` and covered by the new regression test.
- WRITE+ permission is fixed in `9fe41b2` via `requirePermission(request, ["WRITE"])`.
- Dynamic provider discovery remains intentionally deferred for this first status slice. The endpoint is descriptor-based so it stays DB-free and safe for plugin bootstrap; dynamic active-provider wiring belongs with the recall API/provider registry follow-up.

Current bot re-review says previous issues resolved / recommendation merge, and PR checks are green.

PR #49, opened 2026-05-04
feat(openclaw-plugin): adopt Hindsight-inspired auto-recall improvements

SweetSophia:
## Summary

Adopts key improvements from the vectorize-io/hindsight OpenClaw plugin which has a working  hook implementation.

### 1. Use  first (clean user text)
- Hindsight uses  (clean user text without metadata) as the primary query source
- Falls back to  if  is not available
- This ensures the recall query is not polluted with channel metadata envelopes

### 2. Strip channel metadata envelopes
- Removes patterns like 
- Removes  blocks
- Removes  prefixes and session abort hints
- Removes trailing  metadata

### 3. Session pattern filtering
Added three new config options:
- : Array of glob patterns for sessions to skip entirely
- : Array of glob patterns for read-only sessions
- : Boolean to skip recall for stateless sessions (default: true)

### 4. Inflight recall deduplication
- Added  Map to track in-flight recall requests
- Concurrent recalls for the same query are deduplicated
- Prevents redundant API calls when multiple hooks fire simultaneously

## Testing
- All 149 existing tests pass
- TypeScript compiles cleanly

## Summary by Sourcery

Adopt Hindsight-inspired improvements to the OpenClaw Noosphere auto-recall hook to make recall queries cleaner, more efficient, and configurable per session.

New Features:
- Add support for using a clean rawMessage field as the primary source for auto-recall queries, falling back to the prompt when needed.
- Introduce session pattern filtering options to ignore sessions, mark stateless sessions, and optionally skip recall for stateless sessions based on glob-style patterns.

Enhancements:
- Strip channel and metadata envelopes from recall queries and recent turns to avoid polluting recall with transport-level information.
- Deduplicate concurrent recall requests for the same normalized query using an in-memory in-flight recall map to reduce redundant API calls.
- Relax auto-recall eligibility to consider either rawMessage or prompt presence when deciding to run recall.

### DIFF
--- a/openclaw-noosphere-memory/src/auto-recall.ts
+++ b/openclaw-noosphere-memory/src/auto-recall.ts
@@ -1,4 +1,3 @@
-import { createHash } from "node:crypto";
 import { NoosphereRecallRequest, NoosphereRecallResponse, NoosphereSettingsResponse } from "./client.js";
 import {
   clampTimeout,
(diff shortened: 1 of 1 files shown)


SweetSophia:
@codex review

PR #51, opened 2026-05-06
feat(openclaw-plugin): add Noosphere CLI helpers

SweetSophia:
## Summary
- register an OpenClaw `noosphere` CLI command group from the plugin
- add `status`, `doctor`, `logs`, `setup`, and `upgrade` helper commands
- add doctor/status tests and README documentation
- keep Docker/log operations guidance-only so plugin package installation stays within OpenClaw safety policy

## Review follow-up
- require API key + successful authenticated memory status for `openclaw noosphere status` to be OK
- avoid printing `Memory API: OK` when `/api/memory/status` returns `ok: false`
- apply configured timeout to `/api/health` checks
- omit `apiKeyRedacted` when the API key is missing
- reject fractional `--tail` values
- tighten local CLI command typings without adding Commander as a dependency

## Verification
- `npm run lint`
- `npx tsc --noEmit`
- `npm run test:memory` (155/155)
- `DATABASE_URL=postgresql://placeholder@localhost/placeholder npm run build`
- `cd openclaw-noosphere-memory && npm run build && npm pack`
- `openclaw plugins install ./sweetsophia-openclaw-noosphere-memory-0.1.0.tgz --force`
- `openclaw noosphere --help`
- `openclaw noosphere doctor --json`
- `openclaw noosphere status --json`
- `openclaw noosphere logs db --tail 5`

## Notes
OpenClaw blocks third-party plugin packages that import `child_process`; the first CLI implementation attempted direct Docker checks and was correctly rejected by the installer. This version avoids shell execution in the packaged plugin and prints the Docker log command instead.


SweetSophia:
## Bug Found: `registerCli` called before `cli-metadata` guard

**File:** `openclaw-noosphere-memory/src/index.ts`

```typescript
if (typeof api.registerCli === "function") {
  api.registerCli(({ program }: { program: unknown }) => registerNoosphereCli(...), {...});
}
if (api.registrationMode === "cli-metadata") return;  // early return AFTER registerCli call
```

If OpenClaw ever invokes the plugin with `registrationMode === "cli-metadata"` **and** provides `registerCli`, the CLI would be registered before the early return exits — defeating the purpose of the guard. The `cli-metadata` mode exists precisely to skip registration during metadata collection.

**Fix:** Move the `cli-metadata` guard before the `registerCli` call:

```typescript
if (api.registrationMode === "cli-metadata") return;
if (typeof api.registerCli === "function") {
  api.registerCli(({ program }: { program: unknown }) => registerNoosphereCli(...), {...});
}
```

**Severity:** Uncertain — depends on whether OpenClaw ever passes `registrationMode === "cli-metadata"` when `registerCli` is also available. If those two conditions never co-occur in practice, this is a no-op. However, the current ordering is fragile and the guard should logically come first.

PR #60, opened 2026-05-10
fix(security): CSP headers, XSS prevention, Prisma pool leak, rate limiting

SweetSophia:
## Summary
Fixes critical security and infrastructure issues identified in code review.

## Changes
- **CSP & Security Headers**: Adds Content-Security-Policy, X-Frame-Options, X-Content-Type-Options, and Referrer-Policy via Next.js middleware.
- **XSS Prevention**: Adds <code>rehype-sanitize</code> to all <code>ReactMarkdown</code> renderers (article pages + preview tabs) to strip dangerous HTML from user-generated markdown.
- **SVG Upload Security**: Replaces weak regex with HTML entity decoding + comprehensive dangerous pattern detection (scripts, event handlers, javascript: URIs, foreignObject, etc.).
- **Prisma Pool Leak**: Fixes pg.Pool leak in development hot reload by reusing the globally cached pool instance.
- **Rate Limiting**: Adds lightweight in-memory rate limiting for auth endpoints (10/min) and write-heavy API endpoints (30/min).
- **Error Standardization**: Introduces <code>apiError()</code> helper with requestId tracing for easier debugging.

## Files Changed
- <code>src/middleware.ts</code> (new)
- <code>src/lib/rate-limit.ts</code> (new)
- <code>src/lib/api/errors.ts</code> (new)
- <code>src/lib/prisma.ts</code>
- <code>src/lib/uploads.ts</code>
- <code>src/app/wiki/[topicSlug]/[articleSlug]/page.tsx</code>
- <code>src/components/wiki/MarkdownPreviewTabs.tsx</code>
- <code>package.json</code> / <code>package-lock.json</code> (adds <code>rehype-sanitize</code>)
- <code>.gitignore</code>

---
*Part of code review remediation (#60)*

## Summary by Sourcery

Harden security around request handling, user-generated content, and database connections while adding basic rate limiting and standardized API error responses.

New Features:
- Introduce middleware to apply CSP and other security headers across APIs, login, and uploads routes.
- Add a reusable in-memory rate limiting utility and apply it to authentication and write-heavy API endpoints.
- Provide a standardized apiError helper for structured JSON error responses with request IDs for tracing.

Bug Fixes:
- Prevent Prisma connection pool leaks in development by reusing a cached pool instance.
- Strengthen SVG upload validation by decoding HTML entities and detecting a broader set of dangerous patterns before accepting SVG files.

Enhancements:
[...]

PR #64, opened 2026-05-10
fix(perf): cap lint articles, bump tsconfig target to ES2022

SweetSophia:
## Summary
Performance guardrails and modern TypeScript target.

## Changes
- **Cap lint articles**:  now accepts an optional  body field (default 500, hard max 2000) to prevent O(N²) regex processing on large wikis. The query uses  +  for deterministic results.
- **Bump TypeScript target**:  target upgraded from  to . Next.js 16 fully supports this and it unlocks native , , and other modern features without polyfill assumptions.

## Files Changed
- 
- 

---
*Part of code review remediation (#60)*

## Summary by Sourcery

Limit the number of articles processed by the lint API for performance and update the TypeScript compilation target to ES2022.

New Features:
- Allow the lint API to accept an optional maxArticles parameter to control how many articles are inspected per request.

Enhancements:
- Cap linted articles and sort them by most recently updated to avoid excessive processing on large datasets.
- Upgrade the TypeScript compiler target from ES2017 to ES2022 to leverage more modern JavaScript features.

### DIFF
--- a/src/__tests__/api/lint.test.ts
+++ b/src/__tests__/api/lint.test.ts
@@ -1,0 +1,104 @@
+import assert from "node:assert/strict";
+import test from "node:test";
+
+// Test the same parsing logic used in the lint route
+// Constants must match src/app/api/lint/route.ts
+const LINT_MAX_ARTICLES_DEFAULT = 500;
+const LINT_MAX_ARTICLES_HARD_LIMIT = 2000;
+
+function parseMaxArticles(value: unknown): number {
+  const parsed = Number(value);
+  if (Number.isNaN(parsed)) {
+    return LINT_MAX_ARTICLES_DEFAULT;
+  }
+  return Math.min(Math.max(1, parsed), LINT_MAX_ARTICLES_HARD_LIMIT);
+}
+
+test.describe("maxArticles parsing logic", () => {
+  test("uses default when undefined", () => {
+    assert.equal(parseMaxArticles(undefined), 500);
+  });
+
+  test("uses default when NaN", () => {
+    assert.equal(parseMaxArticles(NaN), 500);
+  });
+
+  test("uses default when string 'invalid'", () => {
+    assert.equal(parseMaxArticles("invalid"), 500);
+  });
+
+  test("uses default when object", () => {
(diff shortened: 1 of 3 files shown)


SweetSophia:
## Required Before Merge: 3 Issues

### 1. 🚨 `sync/obsidian/route.ts` — Permission Changed (Behavioral Breaking Change)

The POST handler now requires `ADMIN` permission only, but the old code allowed `WRITE` keys:

```typescript
// BEFORE (WRITE or ADMIN):
if (apiAuth.permissions !== "ADMIN" && apiAuth.permissions !== "WRITE") { ... }

// AFTER (ADMIN only):
const auth = await requirePermission(request, [Permissions.ADMIN]);
```

**Action needed:** Document this in the PR description. Existing WRITE API keys will lose ability to trigger sync.

---

### 2. ⚠️ `graph/route.ts` — Two Queries Instead of One (Potential Regression)

The PR splits one query into two:

```typescript
// OLD: 1 query
const articles = await prisma.article.findMany({
  select: { ..., content: true },
});

// NEW: 2 queries
const articles = await prisma.article.findMany({
  select: { ..., content: false },  // without content
});
const articleContents = await prisma.article.findMany({
  select: { id, slug, content: true },  // content only
});
```

**Action needed:** Benchmark to verify this is actually faster. Two round-trips may be slower than one for small datasets.

---

### 3. ⚠️ `lint.test.ts` — Duplicated Constants

Test file duplicates constants from route:

```typescript
// src/__tests__/api/lint.test.ts:
const LINT_MAX_ARTICLES_DEFAULT = 500;  // DUPLICATED from route!
const LINT_MAX_ARTICLES_HARD_LIMIT = 2000;
```

**Action needed:** Export constants from route and import in test, or create shared `src/lib/constants.ts`:

```typescript
// In route:
export const LINT_MAX_ARTICLES_DEFAULT = 500;

// In test:
import { LINT_MAX_ARTICLES_DEFAULT, LINT_MAX_ARTICLES_HARD_LIMIT } from "@/app/api/lint/route";
```

---

Overall the PR is well-structured. These are just implementation details to address before merge.

PR #70, opened 2026-05-11
fix(db): add Article.updatedAt index for lint query performance

SweetSophia:
## What

Add a PostgreSQL index on `Article.updatedAt` to support the capped `/api/lint` query added in PR #64.

## Why

The lint endpoint orders articles by `updatedAt DESC` with a cap of 2000 rows:

Without an index, PostgreSQL must perform a **full table scan + in-memory sort** on large wikis. The new index allows the planner to use an **index scan** instead, keeping lint fast regardless of wiki size.

## Changes

- `prisma/schema.prisma`: `@@index([updatedAt])` on the `Article` model
- `prisma/migrations/20260511091912_add_article_updatedat_index/migration.sql`: creates `Article_updatedAt_idx`

## Testing

Applied locally via `docker compose exec db psql`:

```
# CREATE INDEX
Index verified: Article_updatedAt_idx exists on Article(updatedAt)
```

---

Addressed the Copilot review comment from PR #64.

## Summary by Sourcery

Add a database index on Article.updatedAt to improve performance of the capped /api/lint query on large wikis.

Bug Fixes:
- Improve /api/lint query performance and avoid full table scans by indexing Article.updatedAt.

Deployment:
- Add Prisma migration to create the Article_updatedAt_idx index in PostgreSQL.

### DIFF
--- a/prisma/migrations/20260511091912_add_article_updatedat_index/migration.sql
+++ b/prisma/migrations/20260511091912_add_article_updatedat_index/migration.sql
@@ -1,4 +1,5 @@
--- Create index on Article.updatedAt to improve lint query performance
--- The /api/lint endpoint orders by updatedAt desc and is now capped at 2000 rows,
--- but without an index this ORDER BY causes a full table scan + sort on large wikis.
-CREATE INDEX "Article_updatedAt_idx" ON "Article"("updatedAt");
+-- Composite index on Article(updatedAt DESC, id ASC) to support the lint query's
+-- ORDER BY [{ updatedAt: 'desc' }, { id: 'asc' }] without a sort step.
+-- The IF NOT EXISTS guard handles environments where the index was applied manually
+-- (e.g. hotfix) but the migration hadn't been recorded in _prisma_migrations yet.
+CREATE INDEX IF NOT EXISTS "Article_updatedAt_idx" ON "Article"("updatedAt" DESC, "id" ASC);
--- a/prisma/migrations/20260511091912_add_article_updatedat_index/migration.sql
+++ b/prisma/migrations/20260511091912_add_article_updatedat_index/migration.sql
@@ -1,0 +1,4 @@
+-- Create index on Article.updatedAt to improve lint query performance
+-- The /api/lint endpoint orders by updatedAt desc and is now capped at 2000 rows,
+-- but without an index this ORDER BY causes a full table scan + sort on large wikis.
+CREATE INDEX "Article_updatedAt_idx" ON "Article"("updatedAt");
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -77,6 +77,7 @@ model Article {
   @@index([status])
   @@index([confidence])
   @@index([deletedAt])
+  @@index([updatedAt])
 }
 
 // ─────────────────────────────────────────────
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -77,7 +77,7 @@ model Article {
   @@index([status])
   @@index([confidence])
   @@index([deletedAt])
-  @@index([updatedAt])
+  @@index([updatedAt(sort: Desc), id(sort: Asc)])
 }
 
 // ─────────────────────────────────────────────


SweetSophia:
## Issue Found: Migration Not Idempotent

The migration uses `CREATE INDEX` without `IF NOT EXISTS`, which will fail if re-run:

```
ERROR: relation "Article_updatedAt_idx" already exists
```

**Fix:** I've created PR #71 to address this:
https://github.com/SweetSophia/noosphere/pull/71

The fix changes:
```sql
-- Before (fails on re-run):
CREATE INDEX "Article_updatedAt_idx" ON "Article"("updatedAt");

-- After (safe for re-runs):
CREATE INDEX IF NOT EXISTS "Article_updatedAt_idx" ON "Article"("updatedAt");
```

Please merge #71 first, then rebase #70, or close #70 in favor of #71.

SweetSophia:
## PR #70 Review

### Overall Assessment
Correct change with a sound premise. Two issues: one **already addressed** (idempotency via #71) and one worth noting.

---

### Issues Found

#### 1. Migration not idempotent ✅ ALREADY ADDRESSED
**Status:** You caught this yourself in #71 — the migration uses  without  and will fail if re-run after a failed/partial deployment.

**Fix applied in #71:**  — correct.

---

#### 2. Index creation holds an  ⚠️ SUGGESTION
 acquires a table lock that blocks all reads and writes for the duration of the index build. On a large wiki with millions of article rows, this could cause a production outage.

**Fix:** Use  instead. This builds the index without holding a lock on the table.



**Trade-off:**  cannot run inside a transaction. Prisma migrations run in transactions by default, so you'd need to wrap this in a separate migration or use  and then run the  step manually. For a one-time operational change this is worthwhile.

---

### Analysis of the Index Itself

**Single-column vs composite:** The lint query is:


A single-column index on  eliminates the  sort step (B-tree handles  via scan direction). PostgreSQL will scan the index in reverse rather than sorting rows in memory.

A composite  would be *theoretically ideal* but adds complexity for minimal gain —  is almost always  (soft-deleted articles are a minority), so PostgreSQL's planner typically ignores the  part and uses the  ordering directly. **The single-column index is the right pragmatic choice.**

**Index naming:**  — follows Prisma's default convention, no issue.

**Schema placement:**  after other Article indexes — correct.

---

### Summary

| Issue | Severity | Status |
|-------|----------|--------|
| Non-idempotent migration | 🔴 CRITICAL | Fixed in #71 ✅ |
| No  | 🟠 HIGH | Suggestion — fix for production deploys |
| Index design | 🟢 | Correct — single-column B-tree on  is appropriate |

**Recommendation:** Merge #71 first, rebase/merge #70 after. Then consider a follow-up to make the migration use  in a separate non-transactional step.

SweetSophia:
## PR #70 Review

### Overall Assessment
Correct change with a sound premise. Two issues: one **already addressed** (idempotency via #71) and one worth noting.

---

### Issues Found

#### 1. Migration not idempotent — ALREADY ADDRESSED
You caught this yourself in #71. The migration uses `CREATE INDEX` without `IF NOT EXISTS` and will fail if re-run after a failed/partial deployment. Fix in #71 (`CREATE INDEX IF NOT EXISTS`) is correct.

---

#### 2. Index creation holds an AccessExclusiveLock — SUGGESTION
`CREATE INDEX` acquires a table lock that blocks all reads and writes for the duration of the index build. On a large wiki this could cause a production outage.

**Fix:** Use `CREATE INDEX CONCURRENTLY` instead:

```sql
CREATE INDEX CONCURRENTLY "Article_updatedAt_idx" ON "Article"("updatedAt");
```

**Trade-off:** `CONCURRENTLY` cannot run inside a transaction. You'd need to run this step manually outside Prisma's migration pipeline. For a one-time operational change on a live table this is the right call.

---

### Analysis of the Index Itself

The lint query is:
```typescript
where: { deletedAt: null },
take: maxArticles,
orderBy: [{ updatedAt: "desc" }, { id: "asc" }],
```

**Single-column vs composite:** A composite `(deletedAt, updatedAt DESC, id ASC)` would be theoretically ideal but adds complexity for minimal gain. Since `deletedAt` is almost always `null` (soft-deleted articles are a minority), PostgreSQL's planner ignores the `deletedAt` part and uses `updatedAt` ordering directly. The single-column B-tree index eliminates the `ORDER BY` sort step entirely by scanning the index in reverse. **This is the correct pragmatic choice.**

**Index naming** (`Article_updatedAt_idx`) — follows Prisma's default convention, no issue.

---

### Summary

| Issue | Severity | Status |
|-------|----------|--------|
| Non-idempotent migration | Critical | Fixed in #71 |
| No CONCURRENTLY | High | Suggestion — fix before production deploy |
| Index design | Low | Correct — single-column B-tree on updatedAt is appropriate |

**Recommendation:** Merge #71 first, then rebase/merge #70. Consider a follow-up to make the migration use `CONCURRENTLY` in a separate non-transactional step before applying to production.

PR #71, opened 2026-05-11
fix(db): make Article updatedAt index migration idempotent

SweetSophia:
## Summary

Fix PR #70 by making the  migration idempotent using .

## Changes

- Update migration to use  instead of 

## Why

If the migration runs twice (e.g., after a failed first attempt or in CI/CD with manual intervention), it would fail with:

```
ERROR: relation "Article_updatedAt_idx" already exists
```

Using  makes it safe for re-runs.

## Related

- PR #70 (base)
- Addresses review feedback on PR #64

## Summary by Sourcery

Bug Fixes:
- Prevent migration failures when the Article_updatedAt_idx index already exists by making the index creation idempotent.


SweetSophia:
Closing as redundant — the IF NOT EXISTS guard was already added to this migration during PR #70 review and is present in master (commit 253b770). The migration SQL on master already uses CREATE INDEX IF NOT EXISTS.

PR #75, opened 2026-05-12
fix(install): preserve runtime env and print bootstrap failures

SweetSophia:
Investigates and fixes the partial-install failure Sophie hit after the installer printed:\n\n```\nApplying database schema and bootstrap data...\n```\n\nRoot cause:\n- The bootstrap command was run under `set -e` before the script captured `0`.\n- If `docker compose run --rm -T init` failed, Bash exited immediately, so the temp log was never printed.\n- Because `.env` was only written after bootstrap success, the DB container/volume could be left running with a generated Postgres password that was not recoverable by rerunning the installer.\n\nFix:\n- Persist `~/.noosphere/.env` with chmod 600 before starting persistent containers.\n- Recover credentials from `.env` on rerun if the OpenClaw secret file was not written yet.\n- Capture bootstrap exit code safely with `cmd || BOOTSTRAP_EXIT=0` so logs are printed.\n- Tell the operator that runtime config is preserved and the installer can be rerun.\n\nValidation:\n- `bash -n install-openclaw.sh`\n- `git diff --check origin/master...HEAD`\n- helper tests for `env_get` and `write_runtime_env`\n- simulated bootstrap failure confirms error log is printed, `.env` is preserved, and OpenClaw secret file is not prematurely written\n- stubbed successful installer path still writes runtime env + OpenClaw secret file

## Summary by Sourcery

Preserve and reuse runtime installer secrets across reruns and ensure bootstrap failures surface their logs to the operator.

New Features:
- Persist a runtime .env file with installer-generated secrets before starting persistent containers so partial installs can be resumed.

Bug Fixes:
- Recover database and bootstrap credentials from the existing runtime .env on rerun to avoid mismatched secrets with already-initialized PostgreSQL volumes.
- Capture bootstrap command failures without exiting the script prematurely and print the captured logs along with guidance to rerun the installer.

Enhancements:
- Extract helper functions for reading key-value pairs from .env files and writing the runtime environment in a single authoritative place.

### DIFF
--- a/install-openclaw.sh
+++ b/install-openclaw.sh
@@ -29,6 +29,37 @@ json_get() {
   JSON_GET_FILE="$file" JSON_GET_KEY="$key" node -e 'const fs=require("fs"); const p=process.env.JSON_GET_FILE; const k=process.env.JSON_GET_KEY; if (!p || !k || !fs.existsSync(p)) process.exit(0); const data=JSON.parse(fs.readFileSync(p,"utf8")); if (typeof data[k] === "string") process.stdout.write(data[k]);'
 }
 
+env_get() {
+  local file="$1"
+  local key="$2"
+  if [ ! -f "$file" ]; then
+    return 0
+  fi
+  awk -v key="$key" '
+    index($0, key "=") == 1 {
+      print substr($0, length(key) + 2)
+      exit
+    }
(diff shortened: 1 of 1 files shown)


SweetSophia:
@codex review

PR #99, opened 2026-05-17
feat: protect admin routes in proxy

SweetSophia:
## Summary
Adds route-level defense-in-depth for wiki admin pages through the existing Next.js 16 `src/proxy.ts` route proxy.

## Behavior
| Path | Unauthenticated | Non-ADMIN |
|------|-----------------|-----------|
| `/wiki/admin/*` | Redirect to `/wiki/login?callbackUrl=...` | Redirect to `/wiki` |
| All other paths | No change | No change |

## Why
- Fixes **CRIT-3** from the security audit.
- Prevents accidental exposure if a future admin page forgets the per-page auth check.
- Keeps API routes excluded from page-level auth decisions; they continue to use `requirePermission()`.

## Implementation notes
- Integrates admin protection into `src/proxy.ts` instead of adding a separate `middleware.ts`.
- Preserves and validates internal wiki `callbackUrl` values after login.
- Adds `next-auth/jwt` type augmentation for the JWT `role`.
- Extends proxy security tests for admin redirects and similarly named non-admin wiki paths.

### DIFF
--- a/src/__tests__/security/proxy.test.ts
+++ b/src/__tests__/security/proxy.test.ts
@@ -5,17 +5,19 @@ import { NextRequest } from "next/server";
 import { proxy } from "@/proxy";
 
 function request(path: string, init: RequestInit = {}) {
+  const headers = new Headers(init.headers);
+  headers.set("x-real-ip", headers.get("x-real-ip") ?? `test-${crypto.randomUUID()}`);
+  const { headers: _headers, signal, ...requestInit } = init;
+
   return new NextRequest(`http://localhost${path}`, {
-    ...init,
-    headers: {
-      "x-real-ip": `test-${crypto.randomUUID()}`,
-      ...(init.headers ?? {}),
-    },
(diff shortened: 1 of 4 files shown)


SweetSophia:
Addressed the valid review feedback in commit `46b8d44`:

- rebased onto current `master`
- moved admin route protection into existing `src/proxy.ts`; removed the separate `src/middleware.ts` entry point from the PR diff
- preserved `callbackUrl` including query string for admin redirects
- updated login to validate and honor same-origin wiki callback URLs after sign-in
- added `next-auth/jwt` JWT typing for `role`
- extended proxy security tests for unauthenticated admin redirects and similarly named non-admin wiki paths

Local verification:
- `npm run test:security` passed: 9/9
- `npm run lint` passed with the existing 3 warnings on master
- `npm run build` passed with the existing Turbopack NFT warning for obsidian sync tracing

PR #100, opened 2026-05-17
fix: limit graph API cross-reference parsing to prevent CPU exhaustion

SweetSophia:
## Summary
The `/api/graph` endpoint previously regex-parsed returned article content to detect cross-references, which made large wiki graphs vulnerable to unnecessary CPU and memory pressure.

## Changes
- Added `contentLimit` query param (default 100): only the first N articles are candidates for cross-reference parsing
- Added `contentMaxBytes` query param (default 50KB, max 50KB): larger article bodies are not returned to Node for cross-reference parsing
- Articles outside these limits still receive **topic** and **tag** edges
- Rebased onto current `master` so rate limiting and unified route auth are preserved
- Replaced repeated `articles.find(...)` cross-reference lookups with precomputed maps
- Added focused tests for graph query parsing, byte-limit behavior, and lookup maps

## Query params updated
| Param | Default | Max | Description |
|-------|---------|-----|-------------|
| `limit` | 100 | 500 | Max articles returned |
| `contentLimit` | 100 | `limit` | Max articles considered for cross-refs |
| `contentMaxBytes` | 51200 | 51200 | Skip parsing articles larger than this |

## Fixes
- **HIGH-1** from security audit: unbounded memory/CPU usage on graph endpoint

## Future work
- Pre-compute cross-reference edges at write time and query `ArticleRelation` directly

### DIFF
--- a/src/__tests__/api/graph.test.ts
+++ b/src/__tests__/api/graph.test.ts
@@ -1,0 +1,89 @@
+import assert from "node:assert/strict";
+import test from "node:test";
+import {
+  buildArticleLookupMaps,
+  GRAPH_CONTENT_MAX_BYTES_DEFAULT,
+  GRAPH_CONTENT_MAX_BYTES_MAX,
+  isContentWithinByteLimit,
+  parseGraphQueryParams,
+} from "@/lib/graph";
+
+test.describe("parseGraphQueryParams", () => {
+  test("uses safe defaults", () => {
+    assert.deepEqual(parseGraphQueryParams(new URLSearchParams()), {
+      limit: 100,
+      contentLimit: 100,
+      contentMaxBytes: GRAPH_CONTENT_MAX_BYTES_DEFAULT,
+    });
+  });
+
+  test("preserves explicit zero values for content parsing controls", () => {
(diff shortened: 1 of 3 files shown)


SweetSophia:
Addressed the valid feedback in `fedacbb`.

- Rebased onto current `master`, preserving the existing graph rate limit and unified auth.
- Moved graph query parsing into tested helpers so explicit `contentLimit=0` and `contentMaxBytes=0` are preserved.
- Added a hard server-side `contentMaxBytes` ceiling of 50KB.
- Avoided returning large article bodies from the main Prisma graph query; content is loaded only for candidate articles, with PostgreSQL returning `NULL` for bodies over the byte cap.
- Replaced repeated `articles.find(...)` cross-reference lookups with slug maps.

Verification:
- `node --import tsx --test src/__tests__/api/graph.test.ts`
- `DATABASE_URL=... npm run test:api`
- `npm run lint` (0 errors; existing 3 warnings remain)
- `DATABASE_URL=... NEXTAUTH_SECRET=... npm run build` (passes; existing Turbopack NFT warning remains)

PR #103, opened 2026-05-17
feat: add Content Security Policy and security headers

SweetSophia:
## Summary
Adds security headers to all routes via Next.js `headers` config in `next.config.ts`.

## Headers added
| Header | Value | Purpose |
|--------|-------|---------|
| `Content-Security-Policy` | See below | XSS/mitigation |
| `X-Frame-Options` | `DENY` | Clickjacking protection |
| `X-Content-Type-Options` | `nosniff` | MIME sniffing protection |
| `Referrer-Policy` | `strict-origin-when-cross-origin` | Privacy |

### CSP directives
```
default-src 'self'
script-src 'self' 'unsafe-eval'
style-src 'self' 'unsafe-inline'
img-src 'self' data: blob:
font-src 'self'
connect-src 'self'
media-src 'self'
object-src 'none'
frame-ancestors 'none'
base-uri 'self'
form-action 'self'
```

## Notes
- `'unsafe-eval'` is required for Next.js runtime
- `'unsafe-inline'` is required for inline styles (Next.js default behavior)
- Uploaded images are served from `/uploads/images/` (same origin, covered by `'self'`)

## Fixes
- **HIGH-7** from security audit: missing CSP headers

### DIFF
--- a/next.config.ts
+++ b/next.config.ts
@@ -74,6 +74,7 @@ const nextConfig: NextConfig = {
             key: "Content-Security-Policy",
             value: [
               "default-src 'self'",
+              "upgrade-insecure-requests",
               scriptSrc.join(" "),
               "style-src 'self' 'unsafe-inline'",
               imgSrc.join(" "),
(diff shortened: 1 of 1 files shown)


SweetSophia:
@codex[agent] review

PR #108, opened 2026-05-19
fix: harden OpenClaw Noosphere plugin scopes

SweetSophia:
## Summary
- enforce caller scopes for memory saves and article creation, including default restrictedTags for scoped WRITE keys
- keep OpenClaw auto-recall gated by local plugin config and resolve clients per agent
- disable shared corpus supplement by default unless explicitly enabled
- validate plugin baseUrl, tighten npm publish tag filters, add Node engines/package README, and fix installer/docs nits including the stale --no-setup reference
- update stale tests for current slugify/recency behavior

## Verification
- npm run lint (0 errors, 3 pre-existing warnings)
- npm run test:security
- DATABASE_URL=postgresql://test:test@localhost:5432/test npm run test:memory
- npm run test:api with local Noosphere env-derived DATABASE_URL
- npm run build with local Noosphere env-derived DATABASE_URL
- cd openclaw-noosphere-memory && npm run typecheck && npm run build && npm pack --dry-run
- git diff --check

## Notes
- The attached review's --no-setup finding was valid only in the NEXTAUTH_SECRET troubleshooting section; this PR removes that stale flag rather than adding a new installer mode.
- Root build still emits the existing Turbopack NFT warning for obsidian sync import tracing; build exits 0.

## Summary by Sourcery

Harden Noosphere memory and article operations around caller scopes while tightening the OpenClaw Noosphere plugin’s configuration, auto-recall behavior, and publishing workflow.

New Features:
- Allow memory save and article creation tools to accept optional restrictedTags that define access scopes.
- Add configuration flag to control registration of the shared corpus supplement in the OpenClaw Noosphere plugin.
- Support resolving the Noosphere auto-recall client per agent invocation based on plugin configuration.

Bug Fixes:
- Ensure scoped WRITE operations default restricted tags to the caller’s allowed scopes and reject writes that exceed those scopes.
- Gate auto-recall behind the plugin’s local configuration so DB settings cannot enable it when disabled locally.
- Restrict topic listing and counts to data visible under the caller’s scopes, including ancestor topics.
- Normalize and validate Noosphere base URLs, rejecting unsafe or malformed values and preferring HTTPS except for loopback hosts.
- Prevent cross-context recall sharing by keying in-flight recall deduplication per client context.
- Add timeout-aware settings() calls for the Noosphere client to avoid hanging auto-recall.

Enhancements:
- Expose secret-detection helper for reuse in article creation and align memory/article input validation around shared utilities.
- Extend memory save pipeline to carry restricted tags through validation, persistence, and revisions.
- Improve OpenClaw Noosphere plugin docs and README around security, corpus usage, and installer instructions, and add Node engines metadata.
- Adjust tests and defaults to reflect current slugification, provider recency, and enabled provider behavior, and switch tests to HTTPS base URLs.

Build:
- Tighten npm publish workflow to only publish version tags for OpenClaw and OpenCode packages and ensure type-checking is enforced in CI.

Deployment:
- Allow overriding the bind address for the Noosphere app service in the Docker Compose template via environment configuration.

Documentation:
- Document new allowDefaultCorpusSupplement configuration and clarify scoped API key behavior and installer usage in OpenClaw setup docs.
- Add a dedicated README for the OpenClaw Noosphere memory plugin describing installation, configuration, tools, and security considerations.

Tests:
- Expand OpenClaw Noosphere plugin tests to cover per-agent auto-recall client resolution, local gating of auto-recall, and corpus supplement behavior.
- Update memory save and wiki tag tests to assert default restrictedTags behavior, scope enforcement, and updated slug handling.


SweetSophia:
## Review Notes: Issues 1-2

### 1. Inconsistent `restrictedTags` DB validation between memory save and article create

`src/lib/memory/api/save.ts:readRestrictedTagsForCaller` validates that the caller has the requested scopes but does **not** verify the tags exist in the `RestrictedScope` table. `src/app/api/articles/route.ts:resolveRestrictedTagsForCaller` does both checks (it queries `prisma.restrictedScope.findMany` to validate existence).

If a memory save request with invalid `restrictedTags` reaches the DB write, it will fail with a Prisma foreign-key error rather than a clean HTTP 400. This should be either:
- **Unified**: extract a shared `validateRestrictedTags(tags[], allowedScopes[])` helper that does both scope-enforcement and DB-existence checks (async, usable in both routes), or
- **Documented**: explicitly document that memory save defers DB validation and that callers should pre-validate tags against the restricted scope registry

### 2. Duplicated scope-enforcement logic

`readRestrictedTagsForCaller` (save.ts:542) and `resolveRestrictedTagsForCaller` (articles/route.ts:382) are nearly identical — both implement: validate array → normalize deduplication → default to caller scopes when none provided → reject unauthorized scopes → return. They differ only in that one does the DB existence check at the end and is async.

These should be consolidated into a shared helper to prevent future drift. If you go with option 1 above, the shared helper would also solve this.

---

**Note**: Issues 3-6 from the review (anonymous topic listing auth change, BIND_ADDRESS default, CI typecheck hardening, `new Date()` in provider test) are lower risk and don't block merge.

PR #112, opened 2026-05-20
feat: add Redis search cache for faster recall

SweetSophia:
## Summary

Add a Redis-backed cache-aside layer for Noosphere recall/search results. Cached repeat queries avoid the PostgreSQL FTS path while Redis remains optional: if `REDIS_URL` is missing or Redis is unavailable, recall falls back to the database.

### Changes

- **Redis client singleton** (`src/lib/cache/redis.ts`) with lazy connection, short connect timeout, offline queue disabled, terminal-client reset, and fail-open behavior when Redis is unavailable
- **Search cache wrapper** (`src/lib/cache/search-cache.ts`) with SHA-256 cache keys, cache versioning, 30s TTL, and stale-write protection
- **Cache integration in `NoosphereProvider.search()`**: checks Redis before DB query and starts best-effort cache population on misses
- **Constant-time invalidation**: article/tag/topic write paths bump the Redis cache version instead of scanning/deleting cache keys
- **Redis services** added to Docker Compose files with healthchecks and persistent volumes
- **Automated coverage**: cache tests are now part of `npm test` via `npm run test:cache`

### Architecture

```
Recall request
  -> Read Redis cache version
  -> Build SHA-256 key from query + filters + scopes + version
  -> Cache hit: return cached results
  -> Cache miss: query PostgreSQL FTS
               -> Best-effort store if cache version is unchanged
               -> Return DB results
```

### Cache Invalidation

Article, tag, topic, import, ingest, answer-save, and wiki write paths invalidate by incrementing the cache version key. Existing cached entries expire by TTL and become unreachable immediately because new searches use the new version in their cache key.

### Verification

- `npm run test:cache` -> 6/6 passing
- `DATABASE_URL='postgresql://test:test@localhost:5432/test' npm test` -> memory 168/168, cache 6/6, API 49/49, security 9/9 passing

### Notes

- Redis is optional; the app still works without `REDIS_URL`.
- The obsolete implementation plan document was removed to avoid stale unsafe examples drifting from the code.


SweetSophia:
@codex review

PR #123, opened 2026-05-24
fix: correct sync conflict detection with writtenHash

SweetSophia:
Addresses review feedback from PR #122.

**Root cause**: canonicalHash uses a fake timestamp so it never matches the disk file hash. Every file was always flagged as a conflict.

**Fix**: Store writtenHash (hash of exact bytes written to disk) in the manifest. Conflict detection compares disk hash against writtenHash.

**Reviewer issues addressed**:
- Sourcery bug_risk: removed unsafe byte-size heuristic
- Gemini high: fixed broken hash comparison
- Kilo warning: preserveLocalChanges always respected

test:memory 177/177, build succeeds. 1 file +25/-28.

## Summary by Sourcery

Correct sync conflict detection by tracking and comparing the hash of the exact bytes last written to disk instead of the canonical content hash.

Bug Fixes:
- Ensure conflict detection compares the on-disk file hash against the stored written hash, preventing every file from being incorrectly flagged as a conflict.
- Respect preserveLocalChanges when conflicts are detected by either backing up local changes or overwriting with database content as configured.

Enhancements:
- Persist a writtenHash in the sync manifest for each article, derived from the exact bytes written to disk, and reuse it on subsequent syncs when files are unchanged.

### DIFF
--- a/src/lib/obsidian-sync/index.ts
+++ b/src/lib/obsidian-sync/index.ts
@@ -143,11 +143,33 @@ async function gitStatusPorcelain(dir: string): Promise<string> {
   });
 }
 
+function normalizeManagedGitPaths(vaultPath: string, managedPaths: string[]): string[] {
+  const safePaths = new Set<string>();
+
+  for (const managedPath of managedPaths) {
+    const safeAbsolutePath = safePath(vaultPath, managedPath);
+    if (!safeAbsolutePath) continue;
+
+    const relativePath = pathRelative(vaultPath, safeAbsolutePath).replace(/\\/g, "/");
+    if (!relativePath || relativePath === "." || relativePath === ".." || relativePath.startsWith("../")) {
+      continue;
+    }
+    if (relativePath === ".git" || relativePath.startsWith(".git/")) {
+      continue;
+    }
+
+    safePaths.add(relativePath);
+  }
+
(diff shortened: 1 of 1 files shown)


SweetSophia:
## Review Comment: Additional Issues Found

### 1. Dead Code - `fileHash` function (line 382-389)
The `fileHash` function is now unused after the refactor. The inline hash computation at lines 595-596 replaced it. Consider removing this dead code in a follow-up cleanup.

### 2. Behavioral Change - `managedPaths` push (line 628)
The condition `if (existingEntry)` was removed, meaning unchanged files are now pushed to `managedPaths` for git staging even when their content wasn't modified. This may stage unchanged files unnecessarily. Please confirm this is intentional.

### 3. Other Issues Not Addressed by This PR
This PR correctly fixes the hash inconsistency bug. However, the following issues from `docs/OBSIDIAN-SYNC-REVIEW.md` are NOT addressed:
- HIGH: Advisory lock fallback race condition
- HIGH: `safePath` trailing slash bypass
- HIGH: Cleanup deletes articles outside filter scope  
- MEDIUM: Git argument injection risk
- MEDIUM: GET endpoint leaks vault path
- LOW: `writeLastRun` writes empty object (line 352)

PR #128, opened 2026-05-26
feat: add sync conflict preferences API

SweetSophia:
## Summary
- add DB-backed singleton SyncConflictPreferences model and migrations
- expose GET/POST /api/sync/conflict-preferences with READ-safe reads and ADMIN-only writes
- validate preserve/overwrite/manual-review defaults and per-direction preferences for noosphere-to-vault and vault-to-noosphere
- use atomic partial upserts to avoid lost updates and enforce bounded POST body reads
- add API-level tests for defaults, validation, public response shape, auth policy constants, and structural body errors

## Verification
- node --import tsx --test src/__tests__/api/sync-conflict-preferences.test.ts (9/9)
- DATABASE_URL=<docker-network noosphere> npm run test:api (66/66)
- DATABASE_URL=<local noosphere> npm run test:memory (177/177)
- DATABASE_URL=<local noosphere> npm run test:cache (6/6)
- DATABASE_URL=<local noosphere> npm run test:security (9/9)
- npm run lint (0 errors, 3 pre-existing warnings)
- DATABASE_URL=<docker-network noosphere> npm run build
- prisma migrate deploy applied via Docker network db:5432
- GitHub checks passed: Analyze actions/javascript-typescript/python, CodeQL, Kilo Code Review, Sourcery review, docker

## Notes
- This stores Phase 2 preferences only. Conflict review UI, reverse scanner, and apply workflow remain later phases.

### DIFF
--- a/prisma/migrations/20260526172000_add_sync_conflict_preferences/migration.sql
+++ b/prisma/migrations/20260526172000_add_sync_conflict_preferences/migration.sql
@@ -1,0 +1,17 @@
+-- CreateTable
+CREATE TABLE "SyncConflictPreferences" (
+    "id" TEXT NOT NULL DEFAULT 'singleton',
+    "defaultBehavior" TEXT NOT NULL DEFAULT 'manual-review',
+    "noosphereToVault" TEXT NOT NULL DEFAULT 'manual-review',
+    "vaultToNoosphere" TEXT NOT NULL DEFAULT 'manual-review',
+    "updatedAt" TIMESTAMP(3) NOT NULL,
+
+    CONSTRAINT "SyncConflictPreferences_pkey" PRIMARY KEY ("id")
+);
+
+ALTER TABLE "SyncConflictPreferences" ADD CONSTRAINT "SyncConflictPreferences_defaultBehavior_check" CHECK ("defaultBehavior" IN ('preserve', 'overwrite', 'manual-review'));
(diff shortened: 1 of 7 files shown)


SweetSophia:
Follow-up pushed in `7c2330b`:

- Reworked `upsertSyncConflictPreferences` to use one atomic Prisma upsert for non-empty partial updates, avoiding read-modify-write races.
- Explicitly sets `updatedAt` in the write path. I also verified the previous path did work through Prisma `@updatedAt`, but this removes ambiguity.
- Replaced the `content-length` guard with strict header validation plus actual body byte-size enforcement before JSON parsing.
- Added tests for malformed, empty, negative, and oversized `content-length` values.

Re-ran focused tests, `test:api`, lint, and build locally.

SweetSophia:
Follow-up pushed in `5cb7902`:

- Removed the redundant `@@unique([id])` from `SyncConflictPreferences`.
- Added a follow-up migration to drop `SyncConflictPreferences_id_key` instead of rewriting the already-applied migration.
- Returned validation errors verbatim so structural body errors no longer read like field errors.
- Added a regression test for the structural body validation response.

Re-ran:
- `node --import tsx --test src/__tests__/api/sync-conflict-preferences.test.ts` (9/9)
- `DATABASE_URL=<docker-network noosphere> npm run test:api` (66/66)
- `npm run lint` (0 errors, same 3 pre-existing warnings)
- `DATABASE_URL=<docker-network noosphere> npm run build`

PR #149, opened 2026-05-27
Fix: Apply scope filter to topic article counts on wiki home page

### LINKED ISSUE #139: [MEDIUM] Topic article counts on wiki home page ignore scope filter
## Information Disclosure

The wiki home page `src/app/wiki/page.tsx` fetches topic article counts WITHOUT applying the restricted article scope filter, while the recent articles list correctly applies it.

```typescript
// src/app/wiki/page.tsx
const topicCounts = await prisma.article.groupBy({
  by: ["topicId"],
  _count: { topicId: true },
  where: { deletedAt: null },  // ← No scope filter!
});
```

Meanwhile `recentArticles` correctly uses `scopeWhere` which applies `buildScopeFilter`.

This leaks information about the existence and rough quantity of restricted articles in each topic. An attacker could infer that a topic contains restricted content based on the count mismatch between what their scoped key can access vs. the actual total.

### Fix Required

Apply `buildScopeFilter(auth.auth.allowedScopes, { deletedAt: null })` to the `topicCounts` query, or use a subquery that filters to only articles the current session/key can access.

Note: The scope filter would need to be applied per-topic, not globally, so the query structure may need to change.

### References

- `src/app/wiki/page.tsx` — wiki home page
- `src/app/api/topics/route.ts` — API correctly filters (for comparison)
- `buildScopeFilter` in `src/lib/api/auth.ts`

SweetSophia:
## Summary
- Apply `buildScopeFilter` to the `topicCounts` groupBy query to prevent information disclosure about restricted articles
- The wiki home page was showing total article counts (including restricted) to all users, while the recent articles list correctly applied scope filtering

## Fix
Changed `src/app/wiki/page.tsx` line 95 from:
```ts
where: { deletedAt: null },
```
to:
```ts
where: scopeWhere,
```

This ensures topic article counts are filtered based on the user's access scopes, matching the existing behavior of the `recentArticles` query.

## Security Impact
Prevents attackers from inferring the existence and rough quantity of restricted articles in each topic by comparing what their scoped key can access vs. the displayed total count.

## Summary by Sourcery

Tighten wiki security and background job robustness by applying scope-aware filtering to topic counts, introducing a Redis-backed sliding-window rate limiter, and improving the local memory scheduler’s concurrency handling, with corresponding documentation and tests.

Bug Fixes:
- Apply user scope filtering to topic article count aggregation on the wiki home page to avoid leaking restricted content statistics.

Enhancements:
- Replace the in-memory rate limiter with a Redis-backed sliding-window implementation that supports shared limits across processes and fails open when Redis is unavailable.
- Add concurrency coordination to the local memory scheduler so concurrent runJob and runDueJobs calls share a single execution per job and reuse the same completion result.

Documentation:
- Revise and expand the Noosphere development guide with stack updates, common commands, deployment instructions, and restricted article behavior.
- Add a detailed implementation plan document for the Redis-backed rate limiter under the superpowers plans documentation.

Tests:
- Add a dedicated rate limiter test suite validating Redis-backed behavior, fail-open semantics, and header-based client keying.
- Extend local memory scheduler tests to cover concurrent runDueJobs and runJob interactions and ensure consistent job execution snapshots.


SweetSophia:
Splitting out the ready security fix into #150.

This PR still contains the broader Redis rate-limiter/scheduler/docs work and currently should stay draft until that scope is finished independently.

PR #175, opened 2026-06-02
fix(api): hide soft-deleted related article targets

SweetSophia:
## Summary
- adds a shared related-article visibility helper that filters soft-deleted targets before serialization/rendering
- applies it to `GET /api/articles`, `PATCH /api/articles/:id` responses, and the wiki article connected-pages panel
- adds focused regression coverage for soft-deleted stale relations and scope filtering

## Semantic anchors
- STRIDE: Information Disclosure, because stale relation rows could expose deleted target metadata
- TDD: focused regression tests around the relation projection invariant
- GitHub Flow + Definition of Done: small branch, verified locally before PR

## Verification
- `node --import tsx --test src/__tests__/api/article-relations-scope.test.ts src/__tests__/api/article-relations.test.ts` — 15/15 pass
- `npm run typecheck` — pass
- `npm run lint` — pass with 2 pre-existing warnings
- `set -a; . ./.env.test; set +a; npm run build` — pass, with existing Turbopack tracing warning
- `node --env-file=.env.test --import tsx --test 'src/__tests__/api/**/*.test.ts'` — 236/237 pass; pre-existing `sync-conflict-review` Node mock failure remains\n- `npm audit --omit=dev --audit-level=high` — no high/critical vulnerabilities; existing moderate `next-auth -> uuid` advisory remains\n\n## Notes\n`npm test` without env still fails on API tests that import Prisma without `DATABASE_URL`; reran API suite with `.env.test` for the meaningful local gate.

## Summary by Sourcery

Ensure related articles never expose soft-deleted targets in API and wiki responses.

Bug Fixes:
- Prevent soft-deleted related article targets from being returned in article API responses and wiki connected-pages views.

Enhancements:
- Introduce a shared helper to filter visible related-article rows using both soft-delete and scope restrictions, and apply it to API and wiki render paths.

Tests:
- Add regression tests to verify that soft-deleted related article targets are filtered out and that existing scope-based filtering remains unchanged.

### DIFF
--- a/src/__tests__/api/article-relations-scope.test.ts
+++ b/src/__tests__/api/article-relations-scope.test.ts
@@ -198,3 +198,20 @@ test("filterVisibleRelatedArticleRows keeps existing scope filtering for live ta
 
   assert.deepEqual(result, [unrestricted, allowed]);
 });
+
+test("filterVisibleRelatedArticleRows short-circuits: soft-deleted targets are hidden from admin callers", () => {
+  // Even an admin ('*') cannot surface a soft-deleted target through the
+  // related-articles panel — soft-delete is an unconditional hide, matching
+  // the DELETE handler's 404-on-deleted contract.
+  const trashed = {
+    target: {
+      id: "a-trashed",
+      restrictedTags: ["financial"],
+      deletedAt: new Date("2024-01-01"),
+    },
+  };
+
+  const result = filterVisibleRelatedArticleRows([trashed], ["*"]);
+
+  assert.deepEqual(result, []);
+});
--- a/src/__tests__/api/article-relations-scope.test.ts
+++ b/src/__tests__/api/article-relations-scope.test.ts
@@ -2,6 +2,7 @@ import assert from "node:assert/strict";
 import test from "node:test";
 import {
   filterAccessibleRelatedTargets,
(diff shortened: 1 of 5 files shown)


SweetSophia:
Pushed a small follow-up on top of this branch (`65fe5d2`, `refactor(relations): tighten filterVisibleRelatedArticleRows type and add short-circuit test`). Three polishes:

1. **Drop `| null` from `restrictedTags` in `RelatedArticleRow`** — the Prisma schema declares `Article.restrictedTags` as non-nullable, so the queried field is always `string[]`. The nullable type propagated an incorrect signal and was defended against with dead `?? []` code.

2. **Drop the `?? []` defensive coercion** in the function body — now dead against the actual Prisma return type.

3. **Add an explicit short-circuit test** — the property "even admin `['*']` cannot surface a soft-deleted target" was previously only implicit in the first test's fixture; the new test names the invariant so it's a regression target.

The generic still requires `target.deletedAt: Date | null` (non-optional), so the type system continues to reject any future caller that forgets to `select: { ..., deletedAt: true }`.

`npm run typecheck` clean · `npm run lint` clean (2 pre-existing warnings) · 12/12 relations tests pass.

This PR is the canonical version — closing #174 as a duplicate. Thanks for picking up the fix!

PR #201, opened 2026-06-21
feat(wiki): add loading, not-found, and error route segments

SweetSophia:
This pull request introduces comprehensive route-level state management for the wiki section, significantly enhancing the perceived performance, reliability, and user experience during navigation and data fetching. By leveraging Next.js App Router conventions, it replaces abrupt transitions and unhandled errors with polished, context-aware feedback.

### Key Functional Improvements

*   **Context-Aware Loading States**: Implements granular loading indicators at the global wiki, topic, and article levels. Users now receive specific textual feedback (e.g., "Loading article…" vs. "Loading topic…") alongside a branded spinner during data fetching, reducing perceived latency.
*   **Resilient Error Handling**: Introduces a dedicated client-side error boundary (`error.tsx`) scoped to the wiki. Instead of crashing the application or showing a generic browser error, unhandled server errors now display a user-friendly recovery UI. It includes a "Try again" action, safe navigation links, and surfaces an error digest ID to facilitate easier bug reporting and debugging.
*   **Refined "Not Found" Experience**: Differentiates between generic missing pages and missing articles by providing targeted messaging. Additionally, a new catch-all route (`[...slug]/page.tsx`) ensures that any deeply nested, invalid `/wiki/*` URL correctly triggers the branded wiki 404 page rather than falling back to the generic app-level 404.

### Technical & Accessibility Highlights

*   **Screen Reader Support**: All loading components are built with accessibility in mind, utilizing `aria-busy="true"`, `aria-live="polite"`, and `role="status"` to ensure assistive technologies gracefully announce background navigation without disrupting the user.
*   **Motion Sensitivity**: The CSS implementation includes a `prefers-reduced-motion` media query to automatically disable the loading spinner animation for users with motion sensitivities.
*   **Design System Integration**: The new UI states (`.route-state`, `.route-loading`) seamlessly inherit existing CSS variables for colors, typography, and shadows. This maintains visual consistency across the application while keeping the stylesheet lightweight and maintainable.

---

This pull request updates the error handling behavior in the wiki section to prevent sensitive error details from being exposed to end users in production.

**Changes:**
* **Conditional Error Logging:** Restricts browser console error logging in the wiki error boundary (`error.tsx`) to non-production environments. 
* **Improved Security/Privacy:** Prevents raw client-side error stacks from appearing in the browser's DevTools for production users, while maintaining detailed error logging for local development and debugging.

### DIFF
--- a/src/app/wiki/[...slug]/page.tsx
+++ b/src/app/wiki/[...slug]/page.tsx
@@ -1,0 +1,9 @@
+import { notFound } from "next/navigation";
+
+/**
+ * Routes unmatched nested /wiki/* URLs through the wiki-styled not-found UI
+ * instead of falling back to the app-level/default 404.
+ */
+export default function UnknownWikiRoutePage() {
+  notFound();
+}
--- a/src/app/wiki/[topicSlug]/[articleSlug]/loading.tsx
+++ b/src/app/wiki/[topicSlug]/[articleSlug]/loading.tsx
@@ -1,0 +1,14 @@
+/**
+ * Loading state for the article page.
+ * Shows a lightweight skeleton matching the article page layout.
+ */
+export default function ArticleLoading() {
+  return (
+    <div className="wiki-content" aria-busy="true" aria-live="polite">
+      <div className="route-loading">
+        <div className="route-loading-spinner" aria-hidden />
+        <p className="route-loading-text">Loading article…</p>
+      </div>
+    </div>
+  );
+}
--- a/src/app/wiki/[topicSlug]/[articleSlug]/loading.tsx
+++ b/src/app/wiki/[topicSlug]/[articleSlug]/loading.tsx
@@ -1,11 +1,11 @@
 /**
  * Loading state for the article page.
- * Shows a lightweight skeleton matching the article page layout.
+ * Shows a branded spinner while the article data is fetched.
  */
 export default function ArticleLoading() {
   return (
     <div className="wiki-content" aria-busy="true" aria-live="polite">
-      <div className="route-loading">
+      <div className="route-loading" role="status">
         <div className="route-loading-spinner" aria-hidden />
         <p className="route-loading-text">Loading article…</p>
       </div>
--- a/src/app/wiki/[topicSlug]/[articleSlug]/not-found.tsx
+++ b/src/app/wiki/[topicSlug]/[articleSlug]/not-found.tsx
@@ -1,0 +1,27 @@
+import Link from "next/link";
+
+/**
+ * Article-specific 404.
+ * Gives a more targeted message than the wiki-level not-found.
+ */
+export default function ArticleNotFound() {
+  return (
+    <div className="wiki-content">
+      <div className="route-state">
+        <p className="route-state-eyebrow">404</p>
+        <h1 className="route-state-title">Article not found</h1>
+        <p className="route-state-description">
+          This article may have been moved, renamed, or deleted. Try searching for it or browse the topic.
+        </p>
+        <div className="route-state-actions">
+          <Link href="/wiki" className="btn btn-primary btn-sm">
+            Back to Noosphere
+          </Link>
+          <Link href="/wiki/search" className="btn btn-secondary btn-sm">
+            Search articles
+          </Link>
+        </div>
+      </div>
+    </div>
+  );
+}
--- a/src/app/wiki/[topicSlug]/loading.tsx
+++ b/src/app/wiki/[topicSlug]/loading.tsx
@@ -1,11 +1,11 @@
 /**
  * Loading state for the topic page.
- * Shows a lightweight skeleton matching the topic page layout.
+ * Shows a branded spinner while the topic data is fetched.
  */
 export default function TopicLoading() {
   return (
     <div className="wiki-content" aria-busy="true" aria-live="polite">
-      <div className="route-loading">
+      <div className="route-loading" role="status">
         <div className="route-loading-spinner" aria-hidden />
         <p className="route-loading-text">Loading topic…</p>
       </div>
--- a/src/app/wiki/[topicSlug]/loading.tsx
+++ b/src/app/wiki/[topicSlug]/loading.tsx
@@ -1,0 +1,14 @@
+/**
+ * Loading state for the topic page.
+ * Shows a lightweight skeleton matching the topic page layout.
+ */
+export default function TopicLoading() {
+  return (
+    <div className="wiki-content" aria-busy="true" aria-live="polite">
+      <div className="route-loading">
+        <div className="route-loading-spinner" aria-hidden />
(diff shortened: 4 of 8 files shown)


SweetSophia:
## In-depth review

Solid implementation of F1 from `REVIEW.md`. The follow-up commit `1cfee0c` already addresses the misleading `server console` comment, the inaccurate `skeleton` JSDoc, and adds `role="status"` for a11y. This review covers what the bots didn't catch.

### What's working well

- **Cascade is correct.** `loading.tsx` and `not-found.tsx` at the wiki level will catch every segment that doesn't define its own. Confirmed by tracing the routing tree — `[topicSlug]/page.tsx:36` calls `notFound()` on unknown topic, which bubbles to `src/app/wiki/not-found.tsx` as designed. `error.tsx` covers the whole tree.
- **CSS tokens all exist.** Verified every variable the new CSS uses (`--border-color`, `--accent-color`, `--text-muted`, `--surface-highlight`, `--surface-elevated`, `--shadow-1`, `--danger-color`, `--radius-xl`, `--font-display`, `--text-color`, `--text-faint`) is defined in `src/app/globals.css`. No token drift.
- **Type-check passes** (`tsc --noEmit` clean against `origin/feat/f1-loading-error-route-segments`).
- **`prefers-reduced-motion` is respected** for the spinner.
- **Error digest display is good** — gives users something to quote in bug reports.

### Findings

**1. (Real) Codex is right: the wiki 404 doesn't fire for unmatched multi-segment URLs.** `src/app/wiki/not-found.tsx`'s JSDoc claims it's "triggered by … unknown /wiki/* routes", but that's only true for 2-segment URLs. A 3+ segment URL with no matching route (e.g. `/wiki/admin/nope`, `/wiki/foo/bar/baz`) hits Next.js's default 404 because there's no `[...slug]/page.tsx` catch-all. The article-level `not-found.tsx` only fires when a matched page calls `notFound()`.

   Concrete cases that fall through to the default 404 right now:
   - `/wiki/admin/anything-not-a-real-subpage`
   - `/wiki/topic/article/extra-segment` (typo or stale link)
   - `/wiki/anything/anything/anything`

   **Suggested fix:** add `src/app/wiki/[...slug]/page.tsx` that just calls `notFound()`. One file, 5 lines. This routes all unknown wiki URLs through the wiki-styled 404. Also update the JSDoc on `not-found.tsx` to say "triggered by `notFound()` calls" — the "unknown /wiki/* routes" claim will then be true.

**2. (Minor) Admin segments don't have their own `error.tsx`.** A thrown error in `src/app/wiki/admin/keys/page.tsx` (or any admin page) shows the same "Unexpected error" UI as a wiki article error. For an admin-only operator, that generic copy is acceptable — but consider whether admin errors should differentiate. Not a blocker for this PR; flag for a follow-up.

**3. (Minor) `useEffect` + `console.error` is a half-measure in production.** The follow-up commit's comment correctly notes that Next.js auto-logs server errors in dev. In production, this `console.error` ships to every visitor's DevTools (effectively a stack-trace leak for any client who opens it). Kilo flagged this. For now it's fine — a follow-up should wire the digest to a real error reporter (Sentry, OTel, even a server action that POSTs to `/api/log`). I have this as a deeper feature in `REVIEW.md §C.2` (webhooks + observability).

**4. (Style) The error boundary only logs `error` once, on mount.** If `error` changes (Next.js can pass a new error instance on `reset()`), the effect re-fires. Good. But if the user clicks "Try again" and the same error happens, the effect fires again with a new `error` reference. Good behavior — just verifying this is intentional.

**5. (Style) The 404 copy says "topic or article" but the wiki-level 404 will also fire for unknown `/wiki/admin/*` URLs.** The eyebrow is just `404` and the title is `Page not found` (correctly generic), but the description says "The topic or article you're looking for may have been moved…" — for an admin typo, that's misleading. Consider generic copy: "This page may have been moved, renamed, or deleted."

**6. (Pollyfill) `color-mix(in srgb, var(--danger-color) 30%, transparent)`** has good support in 2023+ browsers but fails silently (no error, just no border color) in older browsers. Probably fine for this product's target audience; flag for awareness.

**7. (Gap) No tests.** I checked — the repo has no React component test infrastructure (no `*.test.tsx`, no testing-library, no jest/vitest setup, no Playwright). All tests are unit tests of lib + API routes using `node:test`. Adding component tests for these new files is non-trivial infrastructure work. Acceptable to defer — but the test plan in the PR description should be updated to reflect that manual testing is the only option, and the unchecked items are real work the author needs to do before merge.

### Recommendations (in order of priority)

1. **Add the catch-all route** (`src/app/wiki/[...slug]/page.tsx` → `notFound()`). This is the only finding I'd push for in this PR. Codex already suggested it; it's a 5-line addition that closes a real gap.
2. **Soften the 404 description copy** to be page-agnostic ("This page may have been moved…").
3. **Update the JSDoc on `not-found.tsx`** to drop the "unknown /wiki/* routes" claim once #1 lands.
4. **Defer everything else** (admin-specific error boundaries, real error reporter, component tests) to follow-ups per `REVIEW.md`.

### Not a blocker, but worth knowing

The PR introduces the first real "polish" change to the wiki UI per `REVIEW.md`. The pattern is reusable: 3 spinner files, 2 not-found files, 1 error file, 1 CSS block. If/when you add per-skeleton loading for high-traffic pages, the existing `.route-loading` block can stay for fallback — the skeletons would just replace `.route-loading-text` with a layout-shaped block. Clean separation between layout (`.wiki-content`) and route state (`.route-state`/`.route-loading`) is already in place.

LGTM once the catch-all lands. Strong PR.

SweetSophia:
Addressed the main actionable finding from the in-depth review in `75e1f42`:

- Added `src/app/wiki/[...slug]/page.tsx` so unmatched nested `/wiki/*` URLs call `notFound()` and render the wiki-styled 404.
- Updated the wiki 404 JSDoc so it no longer overclaims routing behavior.
- Changed the wiki 404 description to page-generic copy.
- Updated the PR description/test plan to include the catch-all and the latest verification.

Verified with `npm run typecheck` and `DATABASE_URL=... npm run build`; the build output now includes `/wiki/[...slug]`.

SweetSophia:
Final subagent review completed.

Result:
- No blocking or important issues found.
- Catch-all route compatibility looks good; explicit `/wiki` routes remain separate and `/wiki/[...slug]` acts as fallback.
- One low privacy note was fixed in `3136955`: full error objects are now logged only outside production, while users can still report the rendered digest.
- Updated the PR test plan to use `/wiki/foo/bar/baz` for catch-all manual verification, because unauthenticated `/wiki/admin/nope` is intercepted by the existing auth flow.

Verified after the final fix:
- `npm run typecheck`
- `DATABASE_URL=... npm run build`

CI has restarted for `3136955`; Snyk is already green and the remaining checks are pending at the time of this note.

PR #205, opened 2026-06-22
refactor: validate API keys by unique hash

### LINKED ISSUE #193: [LOW] validateApiKey uses findFirst + redundant prefix filter; consider findUnique on keyHash
## Summary
`src/lib/api/keys.ts:51-57` uses `prisma.apiKey.findFirst` with `where: { keyPrefix, keyHash, revokedAt: null }`. Since `keyHash` already has a `@unique` constraint (`prisma/schema.prisma:186`), the prefix filter is redundant defense-in-depth.

## Current code
```ts
const record = await prisma.apiKey.findFirst({
  where: {
    keyPrefix: prefix,
    keyHash: hash,
    revokedAt: null,
  },
});
```

## Notes
- `keyHash String @unique` is a B-tree unique index, so the lookup is a single index probe — **not** a full table scan (verified against `prisma/schema.prisma`).
- The `keyPrefix` filter is harmless but redundant. Keeping it as defense-in-depth against a hypothetical hash collision where the prefix doesn't match is defensible; removing it would simplify intent.
- `findFirst` works fine here, but `findUnique({ where: { keyHash: hash } })` makes the "this is uniquely indexed" intent explicit to readers and to the Prisma type system.

## Recommended fix (optional)
```ts
const record = await prisma.apiKey.findUnique({
  where: { keyHash: hash },
});
if (!record || record.revokedAt !== null) {
  return { valid: false };
}
```

## Severity rationale
Not a security or performance issue (the unique index handles both). This is a clarity/intent refactor. Severity downgraded from High to Low after verifying the schema constraint.

## References
- `src/lib/api/keys.ts:42-80`
- `prisma/schema.prisma:183-198`

SweetSophia:
## API Key Validation Refactor

This PR refactors `validateApiKey` to align with the database schema's unique constraint on `keyHash`, replacing the previous combined lookup with a two-step validation flow.

### Key Changes

**Lookup Strategy (`src/lib/api/keys.ts`)**
- Switched from `findFirst({ where: { keyPrefix, keyHash, revokedAt: null } })` to `findUnique({ where: { keyHash } })`
- Removed reliance on `keyPrefix` as a secondary matching field
- Removed the `revokedAt: null` filter from the database query
- Explicitly checks `record.revokedAt !== null` after the lookup, returning `{ valid: false }` for revoked records

This change ensures the query targets the unique index directly, eliminating unnecessary composite filtering and resolving the schema/query mismatch.

**Test Coverage (`src/__tests__/api/keys.test.ts`)**
- Added a test verifying that active keys are looked up by `keyHash` and that `updateMany` is called with the expected `lastUsedAt` condition
- Added a test verifying that revoked records are rejected after the unique lookup and that no `updateMany` is invoked
- Both tests mock `prisma.apiKey` methods, capture call arguments, and restore originals in a `finally` block

### Functional Impact

- Validation now matches the schema's `keyHash` uniqueness constraint, preventing the query from relying on a non-unique composite predicate
- Revoked keys continue to be rejected, but the rejection now happens after the lookup rather than being baked into the query filter
- The `lastUsedAt` timestamp update logic (previously part of the same patch context) is preserved and verified by the new active-key test

Closes #193.

---

**Summary**

Refactored `validateApiKey` to look up API keys by their unique `keyHash` only, instead of filtering by both `keyPrefix` and `keyHash`.

**Details**

- The lookup query changed from `findFirst` (with combined `keyPrefix` + `keyHash` filter) to `findUnique` using just `keyHash` as the unique key.
- Revocation status is now checked in application code after the lookup (`record.revokedAt !== null`) rather than in the database query.
- The function accepts an optional `apiKeys` client parameter (defaulting to `prisma.apiKey`), enabling dependency injection for testing.
- Added three unit tests covering: successful validation, revoked key rejection, and missing key rejection — all verifying that the unique hash lookup and conditional `lastUsedAt` update behave correctly.

---

### Summary

This change refactors `validateApiKey` to look up API key records by a unique `keyHash` rather than by a combined `keyPrefix` + `keyHash` query. This relies on `keyHash` being uniquely indexed in the database, which enables a single `findUnique` lookup instead of a broader `findFirst` scan.

### Functional changes

- **Lookup simplified to unique hash query**: `validateApiKey` now calls `apiKeys.findUnique({ where: { keyHash: hash } })`. The `keyPrefix` slice and the `revokedAt: null` filter were removed from the database query, and the revocation check is now performed in code after the lookup returns.
- **Dependency-injected Prisma client**: The function now accepts an optional `apiKeys` client parameter (defaulting to `prisma.apiKey`). This narrows the surface to just `findUnique` and `updateMany`, which simplifies testing and decouples the function from the full Prisma client.
- **Return shape unchanged**: The function still returns either `{ valid: true, permissions, keyId, allowedScopes }` or `{ valid: false }`.
- **`lastUsedAt` debounce update preserved**: The atomic, debounced `updateMany` that refreshes `lastUsedAt` at most once per 5 minutes continues to run only when the record is found and not revoked.

### Test updates

The test suite was extended with three new cases that exercise `validateApiKey` against a mock Prisma client:

- **Active key**: Confirms the lookup uses `where: { keyHash: expectedHash }`, returns the expected `keyId` / `permissions` / `allowedScopes`, and that `updateMany` is called with the debounce `OR` condition and a fresh `lastUsedAt` date.
- **Revoked key**: Confirms a record with a non-null `revokedAt` is rejected, and that `updateMany` is **not** called.
- **Missing key hash**: Confirms a `null` lookup result is rejected, and that `updateMany` is **not** called.

Existing tests for `generateApiKey` and `hashApiKey` remain unchanged.

---

## Summary

This PR refactors the API key validation logic to use a unique hash lookup, replacing the previous compound prefix-and-hash query, and makes the data access layer injectable for testability.

### Changes

- **`src/lib/api/keys.ts`**
  - Replaced `findFirst` with a `findUnique` query that looks up the API key solely by its `keyHash`. The key hash is now treated as the unique identifier for validation.
  - Removed the `keyPrefix` slice/match against the stored record; the prefix is no longer required for lookup.
  - Moved the `revokedAt !== null` check out of the database query and into application code so revoked keys can be inspected after the unique hash lookup.
  - Introduced a new `ApiKeyValidationClient` interface (with `findUnique` and `updateMany` methods) and `ApiKeyValidationRecord` type so the persistence layer can be substituted in tests. The `validateApiKey` function now accepts an optional client that defaults to `prisma.apiKey`.
  - Added a code comment documenting that the indexed lookup is intentionally not constant-time and that the same invalid response is returned for both missing and revoked keys to prevent API-key enumeration via response shape.

- **`src/__tests__/api/keys.test.ts`**
  - Added comprehensive test coverage for `validateApiKey`:
    - Verifies that active keys are looked up by unique hash, returned with their permissions/allowed scopes/key ID, and that `lastUsedAt` is updated using a debounce OR clause (`lastUsedAt: null` OR `lastUsedAt < cutoff`).
    - Confirms revoked records are rejected after the unique hash lookup without updating `lastUsedAt`.
    - Confirms missing keys are rejected without updating `lastUsedAt`.
    - Verifies that errors from the lookup and the last-used update are propagated to the caller.
    - Verifies that a key remains valid when the debounce window causes the last-used update to be skipped (`count: 0`).
    - Adds an integration test that creates a real API key in the database, validates it, revokes it, and confirms subsequent validation fails. Test data is cleaned up afterwards.

---

Refactor `validateApiKey` to look up API key records by their unique `keyHash` field instead of filtering on `keyPrefix` and `keyHash` together. The function now uses `findUnique` against the unique hash, then collapses both "record missing" and "record revoked" outcomes to the same `{ valid: false }` response, rather than baking the `revokedAt: null` check into the database query.

A new `ApiKeyValidationClient` interface (with `findUnique` and `updateMany` methods) is introduced, allowing the database client to be injected. This makes the validator testable with a mock client. The `lastUsedAt` debounce update remains, still scoped to never-used and stale records.

The test suite is expanded to cover the new behavior, including:

- Verifying the unique-hash lookup and successful validation updates the `lastUsedAt` timestamp with the correct debounce conditions.
- Rejecting revoked records returned from the hash lookup without updating `lastUsedAt`.
- Rejecting keys that don't exist in the database, again without any update call.
- Propagating errors from the lookup and the last-used update.
- Keeping validation successful when the debounced `updateMany` matches zero rows.
- An end-to-end test that creates a real API key in the database, validates it, revokes it, and confirms the next validation fails.

### DIFF
--- a/src/__tests__/api/keys.test.ts
+++ b/src/__tests__/api/keys.test.ts
@@ -1,7 +1,19 @@
 import assert from "node:assert/strict";
-import test from "node:test";
+import test, { after } from "node:test";
 import crypto from "crypto";
-import { hashApiKey, generateApiKey } from "@/lib/api/keys";
+import { Permissions, type Prisma } from "@prisma/client";
+import { prisma } from "@/lib/prisma";
+import {
+  hashApiKey,
+  generateApiKey,
+  validateApiKey,
+  type ApiKeyValidationClient,
+  type ApiKeyValidationRecord,
+} from "@/lib/api/keys";
+
+after(async () => {
+  await prisma.$disconnect();
+});
 
 test("generateApiKey returns a noo_ prefixed key", () => {
   const key = generateApiKey("test-agent");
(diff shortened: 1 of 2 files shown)


SweetSophia:
## PR #205 Review — `validateApiKey` refactor (commit `71d711e`)

### Verdict
**Approve with comments.** The core refactor is correct and well-scoped. `prisma/schema.prisma:186` confirms `keyHash String @unique`, so `findUnique({ where: { keyHash } })` is the sound lookup. The dependency-injection surface (`ApiKeyValidationClient`) cleanly replaces the monkey-patching the earlier reviewers flagged.

### Semantic anchors invoked
**SSOT** — `keyHash` is now the sole lookup key. **Fowler PEAA / Dependency Injection** — narrow port over Prisma enables deterministic stubs. **SOLID — Interface Segregation** — only the two methods `validateApiKey` actually uses. **DRY / KISS** — single-field WHERE; revoked check moved to a one-line application predicate. **MECE** — see coverage matrix below. **BDD Given-When-Then** — each test reads cleanly. **Meszaros — Stub vs Mock** — active test is a stub; revoked/missing tests use a `updateCalled` flag (mock-style behavior verification) which is fine for the negative assertion. **STRIDE — Information Disclosure** — both missing and revoked collapse to `{ valid: false }`; no enumeration regression.

### Important (should fix)

1. **MECE gap in test coverage** (`src/__tests__/api/keys.test.ts`). The current 3 tests cover (active → valid+update, revoked → invalid, missing → invalid). The truth table has more cells:

   | Path | `findUnique` | `updateMany` | Covered? |
   |---|---|---|---|
   | active | returns record | fires | ✅ |
   | revoked | returns record | does **not** fire | ✅ |
   | missing | returns `null` | does **not** fire | ✅ |
   | DB error on `findUnique` | throws | n/a | ❌ |
   | `updateMany` throws after a valid lookup | returns record | throws | ❌ |
   | debounced (`count: 0`) | returns record | fires, `count: 0` | ❌ |

   The third row is the only one with real production risk: a future contributor could re-add `revokedAt: null` to the SQL WHERE and silently bypass the application check. **Add one integration test** that creates a real `ApiKey`, validates successfully, then `prisma.apiKey.update({ data: { revokedAt: new Date() } })` and asserts the next `validateApiKey(raw)` returns `{ valid: false }` end-to-end. That single test closes the only realistic regression vector.

2. **`as LastUsedUpdateWhere` defeats Prisma's generated types** (`src/__tests__/api/keys.test.ts:75`). The helper type duplicates a shape Prisma already maintains. Drop it and assert against `Prisma.ApiKeyUpdateManyArgs['where']` directly, e.g.:

   ```ts
   assert.equal(updateManyArgs.where?.id, "key-active");
   assert.deepEqual((updateManyArgs.where as any)?.OR?.[0], { lastUsedAt: null });
   assert.ok(
     (updateManyArgs.where as any)?.OR?.[1]?.lastUsedAt?.lt instanceof Date
   );
   ```

   Or export `LastUsedUpdateWhere` as `Prisma.ApiKeyUpdateManyArgs['where'] & { OR: ... }` so the `id` field comes from Prisma.

3. **`ApiKeyClientMock` duplicates `ApiKeyValidationRecord`** (`src/__tests__/api/keys.test.ts:7-13`). The mock inlines the same shape the production type defines. If a field is added to `ApiKeyValidationRecord` tomorrow, the mock silently diverges. Either (a) export `ApiKeyValidationRecord` from `keys.ts` and import it in the test, or (b) move both `ApiKeyValidationRecord` and `ApiKeyValidationClient` into a shared `src/lib/api/keys.types.ts`. The mock interface in the test file should then re-use `ApiKeyValidationClient` directly — no parallel declaration.

4. **Pin the error-propagation contract.** `validateApiKey` currently lets DB errors bubble uncaught. That's a reasonable choice, but it's not pinned by any test. Add one test asserting `validateApiKey` rejects when `findUnique` throws, so a future "wrap in try/catch" change doesn't silently alter the API surface.

### Nits / suggestions

5. **`KEY_PREFIX_LENGTH` is now single-use** (`src/lib/api/keys.ts:6`, used only at `:45` inside `generateApiKey`). The validation-side use is gone. Either inline `8` at the call site, move it inside `generateApiKey`, or add a one-line `// display-only truncation for admin UI` comment to head off future-reader confusion.

6. **The `@@index([keyPrefix])` index** (`prisma/schema.prisma:199`) was previously justified by the validation query; it now serves only admin-list queries (none currently filter by `keyPrefix`). Consider dropping it in a follow-up — minor, out of scope here.

7. **Residual timing-attack surface** (pre-existing, no regression): a SHA-256 indexed lookup leaks hash-existence via microsecond-level timing. Worth a one-line `// NOTE: not constant-time` comment above the `findUnique` call so future security reviewers don't have to re-derive this. Not blocking.

### What's resolved (acknowledging earlier reviews)

- **gemini-code-assist** (`t.mock.method`): superseded — DI removed the monkey-patch entirely.
- **sourcery-ai** (monkey-patching): resolved by the `apiKeys` parameter.
- **sourcery-ai** (`unknown` types): resolved — mocks use `Prisma.ApiKeyFindUniqueArgs` / `Prisma.ApiKeyUpdateManyArgs`.
- **kilo-code-bot** (tautological `deepEqual` at `keys.test.ts:79` in the `0b24c84` diff): resolved in `71d711e` — replaced by granular per-field `assert.equal` + `instanceof Date` checks (lines 76–79).
- **kilo-code-bot** (dead `lastUsedAt: null` in mock returns): resolved in `71d711e` — mock returns now match `ApiKeyValidationRecord` field-for-field.

### Recommended action items (priority order)
1. Add the integration test for the revoke-then-revalidate path.
2. Replace `as LastUsedUpdateWhere` with Prisma's generated `where` type.
3. Reuse `ApiKeyValidationRecord` in the test mock instead of re-declaring the shape.
4. Pin the error-propagation contract with one test.

### Approval recommendation
**Approve** — the refactor is correct and the schema's `@unique` makes the new lookup sound. Items 1–4 are test-quality and type-safety polish; none are merge-blocking.

SweetSophia:
Addressed the review follow-up from https://github.com/SweetSophia/noosphere/pull/205#issuecomment-4770524717 in head `1f223a4`.

What changed:
- Added an end-to-end DB regression test: create a real API key, validate it, revoke it, then assert the same raw key returns `{ valid: false }`.
- Reused exported `ApiKeyValidationClient` / `ApiKeyValidationRecord` types in the tests instead of duplicating the production shape.
- Replaced the local `LastUsedUpdateWhere` duplicate type with `Prisma.ApiKeyUpdateManyArgs["where"]` assertions.
- Pinned current error behavior for `findUnique` failures and `updateMany` failures.
- Covered the debounced `updateMany` `{ count: 0 }` path.
- Added a short note near the indexed lookup about the non-constant-time/pre-existing timing surface and same-shape invalid responses.

Verification before push:
- `node --import tsx --test src/__tests__/api/keys.test.ts` passed, 10/10
- `npm run typecheck` passed
- `npm run lint` passed with the two known pre-existing warnings
- production build passed

I am leaving the PR open for human review and will not merge until explicitly asked.

SweetSophia:
Addressed Kilo's latest valid suggestion on the security note in head `44e5fef`.

Change:
- Split the timing and response-shape statements so the comment no longer implies same-shape invalid responses mitigate DB lookup timing.

Verification after the comment-only cleanup:
- `node --import tsx --test src/__tests__/api/keys.test.ts` passed, 10/10
- `npm run typecheck` passed
- `npm run lint` passed with the two known pre-existing warnings
- production build passed

Still leaving PR #205 open for human review; no merge without explicit go-ahead.

PR #206, opened 2026-06-22
Perf: hoist injected memory regex patterns

### LINKED ISSUE #194: [LOW] Hoist regex patterns out of stripOneInjectedTag hot loop
## Summary
`src/lib/memory/api/save.ts:251` constructs `new RegExp(\`<${tag}\\b[^>]*>\`, "i")` on every call to `stripOneInjectedTag`. Because `stripInjectedMemoryBlocks` (line 237-241) loops while `nextContent.changed`, this RegExp is recompiled once per stripped/nested block.

## Current code
```ts
function stripOneInjectedTag(content, tag) {
  const openPattern = new RegExp(`<${tag}\\b[^>]*>`, "i");     // <-- rebuilt every call
  const openMatch = openPattern.exec(content);
  if (!openMatch) return { content, changed: false };

  const closePattern = new RegExp(`</${tag}>`, "gi");          // <-- rebuilt every call
  const openSearchPattern = new RegExp(`<${tag}\\b[^>]*>`, "gi"); // <-- rebuilt every call
  // ...
}
```

The `INJECTED_MEMORY_BLOCKS` array is module-level constant (`save.ts:22-26`), so the tag set is known at import time. All three regexes per tag can be precomputed once.

## Worst case
- A memory-save body is bounded by `MEMORY_SAVE_LIMITS.maxContentLength = 50_000` (`save.ts:14`).
- A body with 5 nested `<recall>` blocks triggers 5 outer calls × 3 regex compilations = 15 `RegExp` constructions.
- Each `new RegExp` with a literal pattern is ~1–5 µs; the absolute cost is microseconds. But the fix is trivial.

## Recommended fix
Hoist the patterns to a module-level `Map<tag, { open, openGlobal, close }>`:

```ts
const INJECTED_MEMORY_PATTERNS: ReadonlyMap<string, { open: RegExp; openGlobal: RegExp; close: RegExp }> = new Map(
  INJECTED_MEMORY_BLOCKS.map((tag) => [
    tag,
    {
      open: new RegExp(`<${tag}\\b[^>]*>`, "i"),
      openGlobal: new RegExp(`<${tag}\\b[^>]*>`, "gi"),
      close: new RegExp(`</${tag}>`, "gi"),
    },
  ]),
);
```

Then `stripOneInjectedTag` looks up `INJECTED_MEMORY_PATTERNS.get(tag)` and `.exec`s on the stored pattern. (Note: the `closePattern.lastIndex` and `openSearchPattern.lastIndex` writes mutate the stored RegExp. If that's a concern, the loop can create local copies via `new RegExp(pat.source, pat.flags)`.)

## Severity
Low. Bounded by `maxContentLength` and the small fixed `INJECTED_MEMORY_BLOCKS` set (3 tags). The fix is mechanical.

## References
- `src/lib/memory/api/save.ts:228-292`
- `src/lib/memory/api/save.ts:12-26` (limits and block list)

SweetSophia:
## Summary

Hoists injected-memory stripping regexes into a module-level pattern record keyed by the known injected block tags.

- Reuses precompiled `open`, `openGlobal`, and `close` patterns per tag instead of constructing new `RegExp` instances on every call to `stripOneInjectedTag`.
- Uses a typed `Record<InjectedMemoryBlock, InjectedMemoryPatterns>` built from `INJECTED_MEMORY_BLOCKS`, so the helper has no unreachable unknown-tag fallback.
- Resets the shared global regex `lastIndex` values at the start of each strip pass, then continues to set `lastIndex` from the current scan cursor inside the nested-block loop.

## Behavior

- Stripping semantics remain unchanged: matching blocks are replaced with newlines, nested same-type blocks are handled by depth tracking, and `strippedBlocks` accounting is preserved.
- The helper now avoids per-call regex compilation while keeping shared `RegExp` instances safe across repeated calls.
- Cross-package behavior is intentionally not unified in this PR; the published OpenClaw package variant is tracked separately in #207.

## Tests

- `memory save strip helper is stable across repeated calls` covers repeated invocations across `recall`, `hindsight_memories`, and nested `recall` input.
- `memory save strip helper handles interleaved injected tags` covers multiple tag kinds in one input and preserves `strippedBlocks` ordering.
- `memory save strip helper pins unclosed variant behavior` pins current server-save behavior for self-closing/unclosed/whitespace-close variants.

## Follow-ups Filed

- #207: deliberate unification of `stripInjectedMemoryBlocks` across server save and the published OpenClaw package.
- #208: server-side injected-block stripping/canonicalization for `POST /api/articles`.

## Verification

- `node --import tsx --test src/__tests__/memory/api-save.test.ts` passed, 21/21
- `npm run typecheck` passed
- `npm run lint` passed with the two known pre-existing warnings
- production build passed
- `openclaw-noosphere-memory` build + typecheck passed

Closes #194.

## Summary

This PR hoists the regex patterns used by the memory save helper's injected-tag stripper out of the per-call hot path. Previously, `stripOneInjectedTag` constructed fresh `RegExp` objects on every invocation; now it reuses a single set of patterns keyed by tag (`recall`, `hindsight_memories`, `noosphere_auto_recall`), with `lastIndex` reset on the stateful global regexes before each scan. The function also gains stricter typing via an `InjectedMemoryBlock` literal-union type so the `tag` parameter is constrained to the configured set, and a comment notes that tag names are embedded directly in `RegExp` sources so metacharacters must be avoided.

A new TODO marker (referencing #207) is added to flag the ongoing effort to unify the server save helper's strip behavior with the OpenClaw package variant, which currently throws on unclosed blocks while the server version drops the trailing content. A matching TODO is added in the OpenClaw `article-create.ts` file. The handler's comment on unclosed blocks now spells out the rationale (preventing transient recall text from leaking into durable memory) and the deliberate behavioral gap.

Test coverage is expanded to lock in the helper's behavior:

- A "stable across repeated calls" test verifies that global regex state does not leak between successive invocations on different inputs.
- A "handles interleaved injected tags" test covers mixed-tag content with multiple block types in one string.
- A "pins unclosed variant behavior" test pins both the self-closing `<recall/>` form and the whitespace-before-`>` form (`<recall>x</recall >`) to the current server-side semantics.

### Files Changed

- `src/lib/memory/api/save.ts` — Hoist and precompile regex patterns; tighten tag typing; document unclosed-block rationale and the #207 unification TODO.
- `openclaw-noosphere-memory/src/tools/article-create.ts` — Add a TODO marker for the same #207 unification effort.
- `src/__tests__/memory/api-save.test.ts` — Add three new tests covering cross-call stability, interleaved tags, and unclosed-tag handling.


SweetSophia:
Addressed Sourcery's valid review feedback in head `181d68e`.

What changed:
- Replaced the `ReadonlyMap.get(...)` lookup with a typed `Record<InjectedMemoryBlock, InjectedMemoryPatterns>`, removing the unreachable missing-pattern fallback.
- Explicitly resets the shared `openGlobal` and `close` regex `lastIndex` values at the start of `stripOneInjectedTag` before any scan work.
- Kept the repeated-call regression test in place to guard against state leaks across calls/tag types.

Verification after the amendment:
- `node --import tsx --test src/__tests__/memory/api-save.test.ts` passed, 19/19
- `npm run typecheck` passed
- `npm run lint` passed with the two known pre-existing warnings
- production build passed

SweetSophia:
Addressed Kilo's valid PR-description feedback.

The body now describes the current implementation accurately:
- `INJECTED_MEMORY_PATTERNS` is a typed `Record<InjectedMemoryBlock, InjectedMemoryPatterns>`, not a `ReadonlyMap`.
- There is no unknown-tag fallback branch in `stripOneInjectedTag`.
- The description now calls out the explicit `lastIndex` reset for shared global regexes.

No code change was needed for this review item.

SweetSophia:
## PR #206 Review — hoist injected memory regex patterns (commit `181d68e`)

### Verdict
**Approve with comments.** The in-`save.ts` refactor is correct, narrowly scoped, and the regression test pins the cross-call `lastIndex` stability claim. The headline finding is an unaddressed duplicate in the **published** npm package `@sweetsophia/openclaw-noosphere-memory@1.10.1` — must be tracked, not silently fixed here.

### Semantic anchors invoked
**DRY** — duplicate across packages. **SSOT** — which copy is the source of truth? **Hyrum's Law** — published behavior is a de-facto contract. **Postel's Law** — silent truncate on unclosed tag. **Fowler PEAA / Module Pattern** — module-level precomputation. **MECE** — test coverage. **ReDoS** — regex safety. **STRIDE** — security. **YAGNI / KISS** — complexity vs. benefit. **Defensive Programming** — the `as` cast and `lastIndex = 0` reset. **Convention over Configuration** — `Object.fromEntries` vs explicit literal.

---

### 🔴 The headline finding: cross-package DRY violation with Hyrum's-Law exposure

`openclaw-noosphere-memory/src/tools/article-create.ts:306-365` is an independent re-implementation of the strip logic. The PR refactors only `src/lib/memory/api/save.ts:22-45, 247-311`. The two are not byte-equivalent and are not interchangeable.

| Aspect | `src/lib/memory/api/save.ts` (refactored) | `openclaw-noosphere-memory/src/tools/article-create.ts:306-365` (untouched) |
|---|---|---|
| Open regex | `<${tag}\b[^>]*>` (word boundary) | `<${tag}(?=[\s>/])[^>]*>` (lookahead) |
| Close regex | `</${tag}>` (strict) | `</${tag}\s*>` (tolerates whitespace) |
| Unclosed tag | **Silently truncates** (`slice(0, openIdx) + "\n"`) | **Throws** `Error("Unclosed memory block tag: <${tag}>")` |
| Patterns | Precompiled at module load (after PR) | Built per call (3× `new RegExp` per `stripOneInjectedTag`) |
| `lastIndex` reset | Explicit at function entry | None (relies on inner-loop `lastIndex = cursor`) |
| `tag` parameter type | `InjectedMemoryBlock` (narrow) | `string` (wide) |

**Why this matters via Hyrum's Law.** Verified that the openclaw duplicate is published:
- `openclaw-noosphere-memory/package.json` has `"private": false`, `"publishConfig": { "access": "public" }`, `"main": "./dist/index.js"`.
- `npm view @sweetsophia/openclaw-noosphere-memory version` → `1.10.1`.
- `dist/tools/article-create.js` contains the literal `Unclosed memory block tag` string (compiled output, 1 match).
- `dist/tools/article-create.d.ts` declares the public surface including the throw contract.

Downstream agents loading `@sweetsophia/openclaw-noosphere-memory@1.10.1` depend on:
1. The `throw new Error("Unclosed memory block tag: <recall>")` contract — caught and surfaced via `errorResult(error, config)` at `article-create.ts:112`.
2. The whitespace-tolerant close regex: `<recall>foo</recall >` is stripped in the openclaw path but **NOT** in the save.ts path.
3. The per-call `lastIndex` semantics where an unclosed tag halts the entire tool call.

Silently unifying the two in a follow-up would change the published behavior of the npm package for every downstream agent. **This PR must leave them alone with a tracked follow-up.**

**Why this matters from a security standpoint (context for the duplicate, not blocking this PR).** `src/app/api/articles/route.ts:299-310` (the server-side `POST /api/articles`) calls `detectSecretInInputs` but **does not** call `stripInjectedMemoryBlocks`. The `openclaw` client-side strip is the only defense on the article-create path — anyone bypassing the plugin (raw HTTP, custom client, future agent framework) can submit `<recall>…</recall>` content that gets persisted as a "published" wiki article. The duplicate is not redundant; it's **load-bearing for the plugin path**. This deserves its own security ticket, out of scope for PR #206.

### Important (should fix in this PR or as an immediate follow-up)

1. **File a tracking issue, do not unify in this PR.** Title suggestion: "Unify `stripInjectedMemoryBlocks` across `src/lib/memory/api/save.ts` and `openclaw-noosphere-memory/src/tools/article-create.ts` (consider shared internal package)". Tag with `security`, `tech-debt`, `Hyrum's-Law`. Include the six behavioral differences. The follow-up should:
   - First, freeze the openclaw contract explicitly (whitespace-tolerant close, throw-on-unclosed).
   - Then, decide which side is SSOT — recommendation: **openclaw** is the more lenient and more spec-aligned, so propagate that to `save.ts`. But this is a deliberate decision, not a refactor accident.
   - Place a `// TODO(noosphere/unify-strip): see <issue>` in both files linking to the tracking issue, so the next hoist doesn't repeat the divergence.

2. **Strengthen the type-safety of `INJECTED_MEMORY_PATTERNS`.** `src/lib/memory/api/save.ts:36-45` uses `as Record<InjectedMemoryBlock, InjectedMemoryPatterns>`. The cast is unsound in theory: a future refactor that hardcodes a literal object could miss a key and `tsc` would not catch it. Two safer patterns:

   ```ts
   // Option A: typed accumulator (preserves the derived-from-array invariant)
   const built: Partial<Record<InjectedMemoryBlock, InjectedMemoryPatterns>> = {};
   for (const tag of INJECTED_MEMORY_BLOCKS) {
     built[tag] = {
       open: new RegExp(`<${tag}\\b[^>]*>`, "i"),
       openGlobal: new RegExp(`<${tag}\\b[^>]*>`, "gi"),
       close: new RegExp(`</${tag}>`, "gi"),
     };
   }
   const INJECTED_MEMORY_PATTERNS = built as Record<InjectedMemoryBlock, InjectedMemoryPatterns>;
   ```

   Or with **runtime exhaustiveness** (the only truly safe pattern):

   ```ts
   const built = Object.fromEntries(INJECTED_MEMORY_BLOCKS.map((tag) => [tag, { ... }]));
   for (const tag of INJECTED_MEMORY_BLOCKS) {
     if (!built[tag]) throw new Error(`stripInjectedMemoryBlocks: missing pattern for <${tag}>`);
   }
   ```

   This converts a silent compile-time hole into a startup-time `Error`. If you keep the current `as` cast, add a one-line `/* keep derived from INJECTED_MEMORY_BLOCKS */` comment so the next reader doesn't refactor it into a literal.

3. **Unclosed-tag behavior is inconsistent across the two copies (Postel's Law call).** The save.ts version silently truncates on unclosed tag — so `<recall>foo` becomes `"\n"`. The mental walk-through: `stripOneInjectedTag` returns `slice(0, openIdx) + "\n"` when no close is found → the open tag is REMOVED. **The strip succeeds and the residual content is well-defined.** ✓ For a security-relevant strip, **silent truncation is safer**: it can never cause downstream data loss from a stray `<recall>` that a generator left open. The current save.ts behavior is correct, but worth a one-line comment because it's load-bearing for the Hyrum's-Law argument above.

### Nits / suggestions (optional)

4. **The `lastIndex = 0` reset at lines 271-272 is defense-in-depth, not strictly necessary.** `save.ts:282-283` already sets `openSearchPattern.lastIndex = cursor` and `closePattern.lastIndex = cursor` *inside* the `while (true)` loop, *before* `exec`. The only code path where the top reset matters is the early-return at line 275 (`if (!openMatch) return { content, changed: false };`). In that branch, no future `exec` call is gated on stale `lastIndex` because:
   - `patterns.open` is **non-global** (`"i"` flag only) — `exec` ignores `lastIndex` on non-global regexes.
   - The next entry to `stripOneInjectedTag` either re-enters the loop body (where inner sets override) or early-returns again.
   **Keep it** — it's a clear signal to a maintainer that these are stateful `RegExp` objects, and it makes the test's regression claim genuinely meaningful. But don't rely on the test below to "prove" the reset is necessary — it isn't, strictly.

5. **JSDoc the module-level `RegExp` cache.** Two lines near `INJECTED_MEMORY_PATTERNS`: "Reused across calls to avoid per-call allocation; stateful `lastIndex` is reset at the top of `stripOneInjectedTag`." This earns its keep the moment someone copy-pastes these into the openclaw package.

6. **Regex injection from `INJECTED_MEMORY_BLOCKS`.** The array is hardcoded as `["recall", "hindsight_memories", "noosphere_auto_recall"] as const`. All three are safe (no regex metacharacters). Worth a one-line comment near the array: `// Tag names must be safe to embed inside a RegExp.source literal (no regex metacharacters).`

### Test quality assessment

The new test `memory save strip helper is stable across repeated calls` (`src/__tests__/memory/api-save.test.ts:99-116`) makes three sequential `stripInjectedMemoryBlocks` calls:

| Sub-case | Input shape | What it exercises |
|---|---|---|
| `first` | Two `<recall>` blocks in one string | Outer `while (nextContent.changed)` re-entering `stripOneInjectedTag` for the same tag |
| `second` | One `<hindsight_memories>` block, separate call | Tag switch with **different** regex objects |
| `third` | Nested `<recall>` in another call | Nested same-type after a previous unrelated call |

**MECE for the "cross-call `lastIndex` leak" hypothesis: partial.**

✅ **What it catches:**
- Any regression that breaks the recursion structure (`while (true)` / depth logic).
- Any change to the output formatting (the trailing `"\n"` joins).
- Any regression where the close regex accidentally becomes non-global (the `strippedBlocks` array would have a different length).
- A full revert to per-call `new RegExp(...)` AND removal of the inner-loop `lastIndex = cursor` assignments.

❌ **What it MISSES (would strengthen the suite):**
1. **Interleaved tags within a single call** — e.g., `"a <recall>1</recall> b <hindsight_memories>2</hindsight_memories> c <recall>3</recall> d"`. Exposes bugs in the outer `for (const tag of INJECTED_MEMORY_BLOCKS)` loop's invariant: that switching `recall → hindsight_memories → recall` correctly resets state.
2. **`noosphere_auto_recall` direct test.** This tag is exercised only transitively in `noosphere_auto_recall + nested recall` (line 78-88). A direct `<noosphere_auto_recall>x</noosphere_auto_recall>` strip would tighten coverage — this is the **least-tested** of the three tags.
3. **Self-closing variants** `<recall/>` and `<recall />` and whitespace-tolerant close `</recall >`. Pre-existing behavior (today: not stripped because no `</recall>` close). Worth a pinning test so any future "fix" is conscious.
4. **Tight-loop amplification** (10+ sequential calls with varying tags) — smoke test for `lastIndex` drift under a hypothetical V8 bug or future regression.

**Test isolation under `--test-concurrency`:** safe. `node --test` runs each *file* in its own process (separate V8 heap, separate module cache), and tests within a file run sequentially. The precompiled `INJECTED_MEMORY_PATTERNS` are module-level state but are not mutated by anything other than `stripOneInjectedTag` (the function under test). ✓

**BDD Given-When-Then:** each test reads setup → call → assert. The new test follows the same idiom. ✓

### Security review (STRIDE)

- **S / T / I:** the precompiled `RegExp` objects are module-private (not exported, not writable from outside the module). No external surface for regex injection. ✓
- **D (ReDoS):** the new patterns are `O(n)` linear scans with no nested quantifiers or backtracking-prone constructs. `<${tag}\b[^>]*>` and `</${tag}>` are both linear in input size. ✓ Confirmed no ReDoS risk.
- **E (Elevation of Privilege via regex injection from `INJECTED_MEMORY_BLOCKS`):** safe today (hardcoded array, no metacharacters). See nit #6.
- **Evasion:** an attacker who knows the tag list can craft `<recall >` (trailing space), `<recall/>` (self-closing), or `<recall attr="x">` — all of these are matched by `<${tag}\b[^>]*>`. So they are stripped. ✓ No regression vs. pre-PR behavior.
- **Case sensitivity:** open patterns use `i` flag; close patterns use `gi`. ✓ `<RECALL>foo</RECALL>` is stripped.
- **Unclosed tags:** as noted, the save.ts version silently truncates. **The strip succeeds and the residual content is well-defined.** ✓ The throw-on-unclosed in the openclaw version is a *strictness* difference, not a security difference.

### Performance claim verification

Construction went from 3 `new RegExp(...)` per `stripOneInjectedTag` invocation to 0 (precomputed once at module load). **The hoist is real, but the actual perf benefit is negligible at the call frequency here.** `stripInjectedMemoryBlocks` is called **once per `POST /api/save` HTTP request** (`src/lib/memory/api/save.ts:214`). For a save endpoint with `rate-limit: 30 req/min` (`src/app/api/articles/route.ts:194`), that's at most 0.5 calls/sec/node. The hoist is more about hygiene (no per-call allocation, no surprise from regex cache eviction) than measurable throughput. **No benchmark in the repo** — claim is plausible-but-unmeasured. The complexity cost is essentially identical (one extra constant + one `as` cast vs. three local `new RegExp` lines), so this is a wash on YAGNI. Verdict: the hoist is acceptable because it's not more complex.

### What previous reviews caught vs. missed

| Reviewer | Caught | Missed |
|---|---|---|
| **kody-ai** (two passes, "No issues found") | Nothing — wrong tool for behavioral equivalence checking | The duplicate entirely; the whitespace-tolerant close divergence |
| **sourcery-ai** | Flow diagram + file-level change table; suggested `Record<InjectedMemoryBlock, InjectedMemoryPatterns>` instead of `ReadonlyMap` ✅ implemented in `181d68e` | The duplicate; the `as` cast safety concern |
| **kilo-code-bot** | Original description said `ReadonlyMap` + missing-tag fallback; didn't match code → author updated PR description ✅ | The duplicate; the actual semantic content of the new test |
| **Human (this review)** | The duplicate + the `POST /api/articles` server-side security gap | n/a |

The duplicate is a **real blind spot** in the review pipeline: all three bots operate on diffs against the PR's stated base branch. None of them looked outside `src/` to find consumers in `openclaw-noosphere-memory/`. A review checklist item — "if this PR refactors a function with the same name in another workspace package, list the differences" — would catch this class of issue systematically.

1. **File a tracking issue** for the openclaw duplicate. Tag with `security`, `tech-debt`, `Hyrum's-Law`. Do not silently unify in a follow-up; do it as a deliberate, reviewable change with the openclaw contract first frozen.
2. **(Optional, in this PR)** Add 2–3 of the missing test cases — interleaved tags, direct `noosphere_auto_recall`, self-closing variant. Cost: ~30 lines. Benefit: pins behavior so the openclaw unification follow-up has a reference oracle.
3. **(Optional, in this PR)** Replace `as Record<...>` with the runtime-exhaustiveness `for` loop variant. Cost: 6 lines. Benefit: convert a silent compile-time hole into a startup-time error.
4. **(Out of scope, but a real security ticket)** `POST /api/articles` does not strip injected blocks. Either move the strip to the server (canonical) or document that the strip is plugin-only defense-in-depth. This deserves its own issue.
5. **(Out of scope, organizational)** Add a `CODEOWNERS` entry for `openclaw-noosphere-memory/` that requires human review on changes touching `article-create.ts`. Or update the bot review checklist to flag "function renamed or hoisted in one workspace package" cross-references.

### Approval recommendation
**Approve** — the in-scope refactor is correct, narrowly scoped, and ships the regression test it claims to; the headline finding is a real but separate DRY/SSOT/Hyrum's-Law problem that must be tracked, not silently fixed here.

SweetSophia:
Addressed Sophie's review comment in head `ca29b15`.

Validity check:
- Confirmed the duplicate implementation in `openclaw-noosphere-memory/src/tools/article-create.ts` is real and behaviorally different from `src/lib/memory/api/save.ts`.
- Confirmed `@sweetsophia/openclaw-noosphere-memory` is published at `1.10.1`, so unifying that behavior should be a deliberate follow-up, not a silent change in this PR.
- Confirmed `POST /api/articles` currently uses `detectSecretInInputs` but does not canonicalize/strip injected memory blocks server-side.

What I changed in this PR:
- Filed tracking issue #207 for deliberate strip-helper unification across server save and the OpenClaw package.
- Filed tracking issue #208 for server-side injected-block stripping on `POST /api/articles`.
- Added `TODO(noosphere/unify-strip): #207` breadcrumbs in both strip-helper implementations.
- Added comments documenting the module-level regex cache, regex-literal-safe tag-name invariant, and intentional server-save unclosed-block truncation behavior.
- Added focused tests for interleaved injected tags and pinned unclosed/self-closing/whitespace-close variant behavior.

Verification after amendment:
- `node --import tsx --test src/__tests__/memory/api-save.test.ts` passed, 21/21
- `npm run typecheck` passed
- `npm run lint` passed with the two known pre-existing warnings
- production build passed

SweetSophia:
Follow-up: the first amended push touched `openclaw-noosphere-memory/src/tools/article-create.ts`, so the package dist check correctly failed because generated `dist/` was stale.

Fixed in head `2405dc0`:
- Ran `npm run build` in `openclaw-noosphere-memory/`.
- Included the regenerated `dist/tools/article-create.js` and source map.
- Verified `npm run typecheck` inside `openclaw-noosphere-memory/` passes.

The earlier root verification remains valid for the source changes:
- focused memory API save tests: 21/21
- root `npm run typecheck`
- root `npm run lint` with the two known pre-existing warnings
- root production build

SweetSophia:
Final Kilo nit addressed in head after `2405dc0`:

- Changed the tag-name comment from "regex-literal safe" to the more precise "avoid metacharacters."
- Updated the PR body so the Tests section lists all three current tests and the Verification section says 21/21.

Verification after this final wording-only source change:
- `node --import tsx --test src/__tests__/memory/api-save.test.ts` passed, 21/21
- `npm run typecheck` passed
- `npm run lint` passed with the two known pre-existing warnings
- production build passed

SweetSophia:
Addressed Kilo's valid current-head nit in `cf2a138`.

What changed:
- Reworded the tag-name invariant comment from "regex-literal safe" to "avoid metacharacters," which matches the actual risk when tag names are interpolated into `RegExp` sources.

Verification after amendment:
- `node --import tsx --test src/__tests__/memory/api-save.test.ts` passed, 21/21
- `npm run typecheck` passed
- `npm run lint` passed with the two known pre-existing warnings
- production build passed

SweetSophia:
## PR #206 Final Inspection — `cf2a138`

### Verdict
**Approve.** All previously-raised actionable findings are addressed. The PR is ready to merge.

### What changed in this final pass

| Finding from prior review | Status | Location |
|---|---|---|
| File tracking TODO in `src/lib/memory/api/save.ts` | ✅ Added | `save.ts:40` (`TODO(noosphere/unify-strip): #207`) |
| File tracking TODO in `openclaw-noosphere-memory/src/tools/article-create.ts` | ✅ Added | `article-create.ts:307-308` |
| Tracking TODO also present in compiled `dist/` | ✅ Added | `dist/tools/article-create.js:248-249` (clean rebuild) |
| Document the unclosed-tag silent-truncate behavior | ✅ Added | `save.ts:292-294` (inline comment + #207 cross-ref) |
| JSDoc the module-level `RegExp` cache | ✅ Added | `save.ts:37-40` |
| Comment about tag-name regex injection | ✅ Added | `save.ts:23` |
| Add interleaved-tags test (was test gap #1) | ✅ Added | `api-save.test.ts:118-130` |
| Add unclosed-variant pinning test (was test gap #3) | ✅ Added | `api-save.test.ts:132-142` |
| Convert `as Record<...>` to runtime-exhaustiveness check | ⚠️ Not done — mitigated | `save.ts:50` still uses `as`, but the new JSDoc at `:39` ("Keep this derived from `INJECTED_MEMORY_BLOCKS` so every configured tag has patterns") makes the invariant explicit |
| Add tight-loop amplification test | ⏭️ Skipped (was a smoke-test suggestion) | — |
| Address `POST /api/articles` server-side strip gap | ⏭️ Out of scope (separate security ticket) | — |
| Add `CODEOWNERS` entry for `openclaw-noosphere-memory/` | ⏭️ Out of scope (organizational) | — |

### Verification — all green

- **Tests:** `node --import tsx --test src/__tests__/memory/api-save.test.ts` → 21/21 pass, 221ms total. All three new tests (stability, interleaved, unclosed-variant) pass.
- **Typecheck:** exit 0, no errors.
- **Lint:** 2 pre-existing warnings in `actions.ts:99` and `import-applier.ts:358` (`@typescript-eslint/no-unused-vars`), 0 errors. These are the "two known pre-existing warnings" called out in the PR body.
- **Behavior preservation:** the inline `strip*` changes (lines 270-298 in `save.ts`) are byte-identical to the previous commit; the new code is purely additive (comments, TODO, JSDoc) plus a wider `tag: InjectedMemoryBlock` type narrowing. No regression risk for in-scope behavior.
- **Published `dist/`:** `openclaw-noosphere-memory/dist/tools/article-create.js` was rebuilt to include the TODO comment; the rest of the file is byte-identical. Source-map updated as expected.

### On the `as Record<...>` cast (the one open item)

I flagged this in the prior review as a silent compile-time hole. The user did not convert to a runtime-exhaustiveness check. **This is the right call for this PR.** Reasons:
1. The pattern is derived from `INJECTED_MEMORY_BLOCKS` via `Object.fromEntries(...).map(...)`. To introduce a silent miss, a future contributor would have to refactor the derivation into a literal object — and the new JSDoc at line 39 explicitly warns against that.
2. Adding a startup-time `Error` for "missing pattern" is a runtime check on a value that is, today, a build-time invariant. The complexity is not free: it adds a `for` loop, a `throw`, and a second `as` cast. For a three-tag, hand-maintained list, the JSDoc-warning approach is the better trade.
3. If a 4th tag is ever added, the type system still helps: `INJECTED_MEMORY_PATTERNS[tag]` would be typed correctly (narrow union), and the new JSDoc would have to be consciously changed to support a literal-style refactor.

**Verdict on the cast:** acceptable as-is given the new JSDoc mitigation. If the tag list ever grows past ~5 entries or becomes config-driven, revisit and add a runtime check.

### On the new "pins unclosed variant behavior" test

This is the strongest test in the new batch and deserves specific callout. `api-save.test.ts:132-142` pins two divergent behaviors vs. the openclaw duplicate:

1. **Self-closing `<recall/>`** → treated as "found open, no close" → drops remainder with `"\n"`. The save.ts `</${tag}>` strict close regex will not match a self-closing tag (since `/>` is not `>` alone and the regex requires `</recall>` literally with no slash inside the close). Behavior is: open tag is "consumed", but no close match is found, so the unclosed branch fires and the entire content from the open tag onward is replaced with `"\n"`. **Result: `"before \n"`** ✅ correct per the test.

2. **Whitespace-tolerant close `</recall >`** → same behavior: the strict close regex `</${tag}>` does not tolerate the trailing space, so the close match fails, and the unclosed branch fires. **Result: `"before \n"`** ✅ correct per the test.

This means the new test **encodes the save.ts behavior as canonical** — it does NOT match the openclaw package's behavior (which would strip `</recall >` and `</recall\n>` successfully). This is a deliberate pin, and the JSDoc/TODO at `save.ts:40` plus the inline comment at `:292-294` explain the intent. **The unification follow-up (#207) is now well-instrumented**: the openclaw duplicate can be brought into alignment and a new test added on the openclaw side to assert parity. ✅

### Semantic anchors invoked in this final pass
**BDD Given-When-Then** — each new test follows the setup/call/assert pattern. **MECE** — interleaved test fills the outer-loop invariant gap. **Postel's Law** — unclosed-tag silent truncate is now documented and pinned. **Hyrum's Law** — TODO links to #207 ensure the openclaw contract is not silently changed. **Defensive Programming** — JSDoc invariants on the module-level cache. **ReDoS** — re-confirmed, no risk in the new patterns.

### One minor observation (not blocking)
The PR includes changes to `openclaw-noosphere-memory/dist/tools/article-create.js` and the source map. This is a clean rebuild of the published artifact with the TODO comment added. If the project's convention is to gitignore `dist/` and publish via CI, the maintainer may want to either (a) commit only the source change and let CI rebuild, or (b) keep the committed dist as the source of truth and document the policy. Not a blocker for this PR; flagging for awareness.

### Recommended action items (remaining)
1. **Merge PR #206** — the in-scope work is complete, well-tested, and the cross-package duplicate is now properly tracked.
2. **Confirm issue #207 exists** and references the behavioral-difference table from my prior review. If the issue hasn't been opened, file it with the four differences from the table.
3. **Schedule the `POST /api/articles` server-side strip** as a separate security ticket. The openclaw duplicate is load-bearing defense today; the server should be canonical.

### Final approval
**Approve and merge.** The refactor is correct, narrowly scoped, ships a meaningful regression test, and the cross-package duplicate is now tracked via #207 with bidirectional TODOs. The two pre-existing lint warnings are unrelated. The optional `as`-cast runtime check was correctly de-scoped in favor of a JSDoc invariant — this is the right trade for a 3-tag hand-maintained list.

Issue #207, opened 2026-06-22
Unify injected-memory block stripping across server save and OpenClaw package


SweetSophia:
## Context

PR #206 hoisted the injected-memory block regexes in `src/lib/memory/api/save.ts`, but review found a second independent implementation in the published package path:

- `src/lib/memory/api/save.ts`
- `openclaw-noosphere-memory/src/tools/article-create.ts`

The OpenClaw package is published as `@sweetsophia/openclaw-noosphere-memory` and currently has public package metadata (`private: false`, public publish config, `main: ./dist/index.js`). Its behavior is therefore a downstream contract and should not be silently changed as part of the narrow PR #206 performance refactor.

## Known behavioral differences

- Open regex:
  - server save: `<${tag}\\b[^>]*>`
  - OpenClaw package: `<${tag}(?=[\\s>/])[^>]*>`
- Close regex:
  - server save: `</${tag}>`
  - OpenClaw package: `</${tag}\\s*>`
- Unclosed tag:
  - server save: silently truncates the injected block tail
  - OpenClaw package: throws `Error("Unclosed memory block tag: <${tag}>")`
- Pattern allocation:
  - server save: module-level precompiled patterns after PR #206
  - OpenClaw package: builds regexes per call
- Type shape:
  - server save: narrowed `InjectedMemoryBlock`
  - OpenClaw package: `string`

## Desired follow-up

1. Freeze the current OpenClaw package behavior with explicit tests before changing it.
2. Decide the single source of truth for injected-memory stripping.
3. Extract or share one implementation across both paths, preserving or intentionally changing documented behavior.
4. Update package docs/changelog if the published OpenClaw behavior changes.

## Notes

This is intentionally tracked outside PR #206 because the duplicate includes published package behavior and Hyrum's-Law risk.

SweetSophia:
Follow-up confirmed after PR #206 merge (squash commit cda733f). This issue already contains the behavioral-difference table from the review: open regex, close regex, unclosed-tag behavior, pattern allocation, and type shape. PR #206 added bidirectional TODOs pointing here from the server save helper and the published OpenClaw package source/dist path.

SweetSophia:
[...]

PR #209, opened 2026-06-22
refactor(memory): share injected block stripper

SweetSophia:
## Summary

Unifies the duplicated injected-memory stripping logic into a single, mode-configurable implementation, eliminating the drift risk that had allowed the two existing copies to evolve differently.

### New shared module: `openclaw-noosphere-memory/src/injected-memory-strip.ts`

Introduces a single `stripInjectedMemoryBlocks(content, mode)` function that supersedes both prior copies. The module is the canonical source of truth and is also exposed as the package entry point `@sweetsophia/openclaw-noosphere-memory/injected-memory-strip`.

Key design points the new code makes explicit:

- **Block catalog is centralized.** A single `INJECTED_MEMORY_BLOCKS` tuple (`recall`, `hindsight_memories`, `noosphere_auto_recall`) drives both consumer paths, with a derived `InjectedMemoryBlock` type and a comment warning that tag names are embedded directly into RegExp sources.
- **Two named strip modes replace hand-rolled variants.** The mode parameter is a literal union (`"openclaw-article-create" | "server-save"`), with `OPENCLAW_ARTICLE_CREATE_STRIP_MODE` and `SERVER_MEMORY_SAVE_STRIP_MODE` exported as mode constants. A `STRIP_MODE_CONFIGS` table (satisfied against `InjectedMemoryStripModeConfig`) declares three independent knobs per mode:
  - `open`: `"tag-boundary"` (whitespace, `>`, or `/` after the tag) vs `"word-boundary"` (`\b`).
  - `closeAllowsWhitespace`: enables the lenient `</tag\s*>` close pattern.
  - `unclosed`: `"throw"` vs `"truncate-tail"`, deciding whether an unclosed block aborts the call or silently drops the rest of the content.
- **Patterns are prebuilt per mode and reused** via `INJECTED_MEMORY_PATTERN_SETS` so the stateful `lastIndex` reset that the old code already relied on is preserved without per-call allocation.
- **Stateful regexes are reset on every entry** (`patterns.openGlobal.lastIndex = 0; patterns.close.lastIndex = 0;`) and after each `exec` advancement, removing a latent risk when the same helper is invoked repeatedly with the same mode.

### `openclaw-noosphere-memory/src/tools/article-create.ts`

- Removes the local `INJECTED_MEMORY_BLOCKS`, `stripInjectedMemoryBlocks`, and `stripOneInjectedTag` definitions and the `#207` TODO marker.
- Imports the shared helper and replaces the local call site with a single call that passes `OPENCLAW_ARTICLE_CREATE_STRIP_MODE` implicitly through the default.
- Behavior preserved: tag-boundary opens, whitespace-allowed close tags (`</recall >`), and throw-on-unclosed semantics (including the self-closing `<recall/>` form that the matcher treats as unclosed).

### `src/lib/memory/api/save.ts`

- Removes the local `INJECTED_MEMORY_BLOCKS`, `InjectedMemoryBlock`/`InjectedMemoryPatterns` types, the `INJECTED_MEMORY_PATTERNS` table, and the local `stripInjectedMemoryBlocks`/`stripOneInjectedTag` implementations.
- Imports `stripInjectedMemoryBlocksShared` and `SERVER_MEMORY_SAVE_STRIP_MODE` from the shared module.
- The exported `stripInjectedMemoryBlocks(content)` (which other server code already calls) is now a one-line delegation to the shared helper in `SERVER_MEMORY_SAVE_STRIP_MODE`, so the existing public signature is unchanged.
- Behavior preserved: word-boundary opens, strict close tags, and tail-truncation for malformed blocks (so transient recall text is never persisted).

### New tests: `src/__tests__/memory/injected-memory-strip.test.ts`

Adds three behavior-level tests that import directly from the shared module source (not the published `dist/`) to freeze the two distinct contracts:

1. **OpenClaw package contract** — verifies that `</recall >` (whitespace before `>`) is stripped in the article-create mode, and that both `<recall/>` (self-closing) and unclosed `<recall>hidden` throw `Unclosed memory block tag: <recall>`.
2. **Server-save contract** — verifies that `<recall/>` produces tail truncation (`"before \n"`), and that whitespace-padded `</recall >` is treated as malformed under the strict-close mode and likewise truncates the tail.
3. **Cross-mode isolation** — runs the helper with both modes in sequence against different inputs and asserts that prebuilt regex state from one call does not leak into the other, and that multiple adjacent blocks (`<recall>first</recall> ... <recall>second</recall>`) are all stripped and recorded.

Together these tests lock the divergent "throw vs. truncate" and "lenient vs. strict close" behaviors that previously lived in two separate code paths.

### Why this matters

The previous duplication silently preserved two different contracts — the OpenClaw package refuses to publish articles containing malformed recall blocks, while the server save path silently strips the tail to keep durable memory clean. Consolidating them into one mode-driven implementation makes that divergence deliberate (it is a per-mode config row rather than a fork), removes the `#207` TODOs that flagged the duplication, and gives both call sites a single, test-covered source of truth.

---

## Summary

This PR introduces a new shared package, **`@sweetsophia/noosphere-injected-memory`**, that centralizes the logic for stripping transient injected-memory blocks (e.g. `<recall>`, `<hindsight_memories>`, `<noosphere_auto_recall>`) from agent-authored content before it is persisted. It unifies the duplicated implementations that previously lived in the OpenClaw plugin and the Noosphere server, while preserving each call site's existing behavior.

### New shared package: `noosphere-injected-memory/`
- Exports `stripInjectedMemoryBlocks(content, mode)` and two mode constants:
  - `OPENCLAW_ARTICLE_CREATE_STRIP_MODE` — matches tag boundaries, allows whitespace before the closing tag, and **throws** on unclosed blocks (preserves current OpenClaw package contract).
  - `SERVER_MEMORY_SAVE_STRIP_MODE` — uses word boundaries, requires tight closing tags, and **truncates the tail** on unclosed blocks (preserves current server save contract).
- Builds per-mode regex pattern sets once and reuses them across calls, resetting stateful global regexes before scanning.

### OpenClaw plugin (`openclaw-noosphere-memory/src/tools/article-create.ts`)
- Removes the local `INJECTED_MEMORY_BLOCKS`, `stripInjectedMemoryBlocks`, and `stripOneInjectedTag` implementations.
- Now imports the helper from `@sweetsophia/noosphere-injected-memory` and calls it with `OPENCLAW_ARTICLE_CREATE_STRIP_MODE`.
- The `TODO(noosphere/unify-strip): #207` note is no longer needed.

### Noosphere server (`src/lib/memory/api/save.ts`)
- Removes the local `INJECTED_MEMORY_BLOCKS`, `INJECTED_MEMORY_PATTERNS`, and `stripOneInjectedTag` implementations.
- The exported `stripInjectedMemoryBlocks(content)` now delegates to the shared helper with `SERVER_MEMORY_SAVE_STRIP_MODE`, so all internal callers see identical behavior.

### Tests (`src/__tests__/memory/injected-memory-strip.test.ts`)
- New test suite verifies that the shared helper preserves the previously divergent behaviors of the OpenClaw package (throws on unclosed/malformed blocks) and the server (truncates tail, treats malformed/empty self-closing blocks as removed).
- Adds a state-isolation test ensuring no leaks between calls/modes.
- Asserts agreement between the package's `dist` entry point and its `src/index.ts` source.

### Build / release pipeline
- `docker-publish.yml`: triggers on changes in `noosphere-injected-memory/**` and adds a CI step that rebuilds the package and fails if the committed `dist/` is out of date.
- `npm-publish.yml`: includes the new package in the publish matrix so it is published alongside the other adapter packages.
- `eslint.config.mjs`: ignores the new package's `dist/` directory alongside the existing ignores.

## Impact

- The Noosphere server and agent plugins no longer depend on each other to strip injected memory blocks; both depend on a single adapter-neutral package.
- Each call site retains its previous behavior exactly, avoiding silent contract changes for the published OpenClaw package.
- The strip logic is now testable from a single location, and CI guards ensure the built `dist` artifact stays in sync with the source.


SweetSophia:
## PR #209 Review — share injected block stripper (commit `49878ea`)

### Verdict
**Request changes.** The refactor's mechanics are sound and behavior is preserved, but the architectural decisions create **real debt that should be paid before merge, not after Hyrum's Law locks it in.**

### Semantic anchors invoked
**DRY** ✓ · **SSOT** ✓ · **SOLID — Dependency Inversion Principle** ✗ · **Law of Demeter** ✗ · **Clean Architecture** ✗ (layering inversion) · **Hyrum's Law** ✗ (new entry point becomes public contract) · **Principle of Least Astonishment** ✗ (default mode throws) · **YAGNI / KISS** — defensible · **Postel's Law** — default should be the least destructive · **Package by Feature vs Package by Layer** — wrong layer · **MECE** — test coverage has gaps · **Defensive Programming** — default is the strictest, not the safest · **Fowler PEAA** — registry of modes is fine

---

### 🔴 The headline finding: architectural inversion (your biggest error)

`src/lib/memory/api/save.ts:14` now imports from `@sweetsophia/openclaw-noosphere-memory/injected-memory-strip`. The Next.js server is now structurally coupled to a **published client plugin**.

**Why this is a layering violation** (verifiable against three independent anchors):

- **SOLID-DIP:** the high-level server policy (memory-save validation) now depends on a low-level concrete artifact (an OpenClaw plugin's compiled `dist/`). The proper resolution is "depend on abstractions; both should depend on them" — but this PR has both call sites depending on the same concrete module that lives in one of them. That's shared concreteness, not DIP.
- **Law of Demeter:** `save.ts` reaches through `@sweetsophia/openclaw-noosphere-memory` into its `injected-memory-strip` subpath. The server is talking to `OpenClawPlugin.injectedMemoryStrip`, not to `injectedMemoryStrip` directly. Two-dot imports are a Demeter smell.
- **Clean Architecture:** the openclaw package is a *plugin adapter*. Adapters belong at the outermost ring, with policies depending inward. This PR reverses that direction.

**Quantified coupling cost** — by depending on `file:./openclaw-noosphere-memory` the server inherits:

| Inherited concern | Source |
|---|---|
| Build artifact (`dist/`) freshness | committed but not server-built |
| `engines.node >= 22` | `package.json:31-33` |
| `openclaw.plugin.json` plugin metadata | irrelevant to server |
| `openclaw.extensions[0]` pointer | irrelevant to server |
| `compat.pluginApi` constraint | irrelevant to server |
| `files: ["dist/", "src/", "openclaw.plugin.json", ...]` | ships plugin metadata to `node_modules/` |
| Version cadence coupled to plugin tag | `npm-publish.yml` gates on plugin tag |
| Future rename/restructuring inside the plugin | breaks server |

**Recommended path: extract `@sweetsophia/noosphere-injected-memory`** (~30 minutes of plumbing).
```
noosphere-injected-memory/
  src/injected-memory-strip.ts   (mode table + helper)
  package.json                   (exports only ".", no openclaw/*, no plugin metadata)
```
Both the openclaw plugin and the server `file:`-depend on this neutral package. Plugin `package.json`/`engines` stay plugin-shaped; server stays server-shaped. Versioning is independent.

**If you must accept the inversion in this PR:** open issue `#210` (extract neutral package) and reference it from both files. Don't merge without a tracked decision — undocumented layering violations are how long-term debt accumulates.

---

### 🔴 F2 — Default mode is the strictest behavior (the footgun)

`injected-memory-strip.ts:51`:
```ts
mode: InjectedMemoryStripMode = OPENCLAW_ARTICLE_CREATE_STRIP_MODE,
```
A casual caller who imports the module and forgets the `mode` arg gets **throws** on `<recall>x`. Principle of Least Astonishment says defaults should be the *least surprising* behavior — here that's the LEAST destructive (truncate-tail).

**Fix (pick one):**
- **Required parameter** (recommended): `mode: InjectedMemoryStripMode` with no default. Forces explicit choice at every call site. The server's `save.ts:230` already passes `SERVER_MEMORY_SAVE_STRIP_MODE`; the openclaw call at `article-create.ts:128-130` would add `OPENCLAW_ARTICLE_CREATE_STRIP_MODE` (one-line change).
- **Default = truncate-tail**: aligns with Postel's Law. Silently dropping half a document is less surprising than a thrown 500.

The accompanying test `injected-memory-strip.test.ts` passes modes explicitly — there is no test of the default-mode path, which is exactly the footgun scenario.

---

### 🔴 F3 — Public API surface added without a version bump

`openclaw-noosphere-memory/package.json:3` is still `1.10.1`. The new `./injected-memory-strip` subpath will be published as part of `1.10.1` on the next npm publish. Per SemVer this should be a **minor bump to `1.11.0`** (additive new entry point; existing `.` export unchanged; `noosphere_article_create` behavior byte-for-byte preserved).

**Why this matters:** once `1.10.1` ships the new entry point, downstream consumers who import it lock the maintainers into preserving:
1. The mode string union (`"openclaw-article-create" | "server-save"`) — renaming is breaking.
2. The error message `"Unclosed memory block tag: <${tag}>"` (line 117) — anyone matching on it gets pinned.
3. The return shape `{ content: string; strippedBlocks: string[] }`.
4. The stripped-block ordering (outer-loop over `INJECTED_MEMORY_BLOCKS`, inner while-loop).

That contract should be advertised as `1.11.0`, not buried in a patch release.

**Action:** bump to `1.11.0` in both `openclaw-noosphere-memory/package.json` and `openclaw-noosphere-memory/package-lock.json` before merge. Run `scripts/sync-version.mjs` if the project convention requires it.

---

### 🟡 F4 — `as Record<...>` cast (now in 2 places)

The `satisfies Record<InjectedMemoryStripMode, InjectedMemoryStripModeConfig>` at line 37 is a real improvement — it makes adding a mode a compile-time error if you forget to give it a config. **But** the `satisfies` only protects the outer mapping; it doesn't transitively narrow the `Object.fromEntries` result at lines 47 and 89. The cast is still unsound if a tag is added to `INJECTED_MEMORY_BLOCKS` without a corresponding pattern entry.

A small TypeScript helper closes the hole without a runtime check:
```ts
const fromEntries = <K extends string, V>(entries: Array<[K, V]>): Record<K, V> =>
  Object.fromEntries(entries) as Record<K, V>;
```
This makes the cast local to one helper. Optional — current state is acceptable but not airtight.

---

### 🟡 F5 — No CI step verifies `dist/` freshness at PR time

Verified by reading `.github/workflows/`:
- `ci.yml` runs `lint` + `typecheck` only.
- `docker-publish.yml:64` and `npm-publish.yml:103,171,241,311` run `git diff --exit-code -- dist`, but **only at release/publish time**.

**Implication:** a PR that changes `openclaw-noosphere-memory/src/injected-memory-strip.ts` will pass PR-time CI without rebuilding. It can be merged and break the *next* Docker build or npm publish.

**Cheap mitigation:** add a `prebuild` script to the server's `package.json:7`:
```json
"prebuild": "cd openclaw-noosphere-memory && npm ci && npm run build && git diff --exit-code -- dist",
"build": "prisma generate && next build"
```
Or add a new matrix job in `ci.yml` that mirrors `npm-publish.yml:99-103` for the three plugins.

---

### 🟡 F6 — Tests import via package path, exercising `dist/` not `src/`

`src/__tests__/memory/injected-memory-strip.test.ts:8` imports via `@sweetsophia/openclaw-noosphere-memory/injected-memory-strip`, which the exports map routes to `dist/injected-memory-strip.js`. Implications:
- The test is implicitly a **dist-freshness check** as well as a behavioral test.
- A source edit without `tsc -p build-tsconfig.json` leaves the test running against stale dist.
- A developer working in `openclaw-noosphere-memory/src/injected-memory-strip.ts` and running server tests sees passing tests against the wrong code.

**Mitigation:** add a parallel source-level test:
```ts
// src/__tests__/memory/injected-memory-strip-source.test.ts
import { stripInjectedMemoryBlocks } from "../../../openclaw-noosphere-memory/src/injected-memory-strip.ts";
```
The server's `tsconfig.json:33` already excludes `openclaw-noosphere-memory`, so the source import resolves at runtime via `tsx` (which `npm run test:memory` already uses). Either add this or document the dist-freshness assumption loudly.

---

### 🟢 F7 — Test coverage is not MECE

The 3 new tests cover: (a) openclaw malformed-block, (b) server-save malformed-block, (c) cross-mode state isolation. MECE would mean ~6-8 small tests. The most important gap is **the default-mode footgun itself is untested**:

```ts
test("shared strip helper default mode throws on unclosed (the footgun)", () => {
  assert.throws(() => stripInjectedMemoryBlocks("before <recall>x after"));
});
```
If the team adopts the "make `mode` required" fix (F2), this test becomes invalid and should be dropped.

Secondary gaps (covered transitively by `api-save.test.ts` via the server wrapper):
- Multiple distinct tags in one call (`api-save.test.ts:118-130` covers via wrapper)
- `noosphere_auto_recall` direct (`api-save.test.ts:78-88` covers via wrapper)

---

### 🟢 F8 — Mode abstraction is defensible

Three knobs (`open`, `closeAllowsWhitespace`, `unclosed`) for two modes. YAGNI purists would say: have two named functions, no table. KISS purists would say: keep the table, it's six lines. **This reviewer finds the current shape defensible:**
- The three knobs are *independent* (`open` doesn't constrain `closeAllowsWhitespace`), so the table compresses 8 possible configs to 2 actual entries cleanly.
- Pre-compiling patterns per mode (lines 42-47) is a measurable win and the abstraction makes it natural.
- The `satisfies` guarantee at line 37 makes adding a third mode cheap and safe.

The `STRIP_MODE_CONFIGS` keys and the exported constants could derive from `Object.keys(...)` for single-source-of-truth, but that's not worth the dance for 2 modes. Keep as-is.

---

### What CI caught vs. missed

**Caught:**
- `ci.yml` lint+typecheck: import resolves, `satisfies` validates, ESLint clean.
- `package-check`: package builds, dist committed and unchanged post-build.
- `docker-publish.yml`: rebuilds dist and diff-checks before Docker build.
- Sourcery, Kilo: orthogonal to architectural concerns.

**Missed by everything:**
- 🔴 The architectural inversion (F1). `npm pack --dry-run` surfaces the package's `files: ["dist/", "src/", "openclaw.plugin.json", ...]` — should have prompted "why is the server depending on a plugin that ships plugin metadata?"
- 🔴 The default-mode footgun (F2). No test exercises the no-arg path.
- 🔴 Version not bumped (F3).
- 🟡 Dist-vs-source test gap (F6).
- 🟡 No CI guard for `dist/` freshness during PR time (F5).

---

### Errors you made — direct assessment

You asked "let us see if I made errors." Direct answers:

1. **You put the shared module in the wrong package.** It should live in a neutral package, not in the openclaw plugin. This is the single biggest mistake and it cascades into F2, F3, F5, F6.
2. **You kept a default on `mode` and chose the wrong default.** Making `mode` required is one keystroke per call site and removes an entire class of runtime surprise.
3. **You didn't bump the version when adding new public surface.** `1.10.1 → 1.11.0` should happen automatically on any new entry point.
4. **You tested against `dist/`, not source.** This makes the test a dist-freshness check, which is what you want a separate test for, not what you want your behavioral tests to be.

The good news: each of these is mechanical to fix. The refactor's *design* (mode-driven single source of truth, `satisfies` for compile-time exhaustiveness, precompiled patterns per mode) is genuinely well-engineered. The architectural placement and the small ergonomic mistakes around it are what need attention.

---

1. **(F1)** Decide: extract `@sweetsophia/noosphere-injected-memory` (recommended) OR file `#210` and reference from both files.
2. **(F2)** Make `mode` required at `injected-memory-strip.ts:51`. Update `article-create.ts:128-130` to pass `OPENCLAW_ARTICLE_CREATE_STRIP_MODE` explicitly.
3. **(F3)** Bump `openclaw-noosphere-memory` to `1.11.0` in `package.json` and `package-lock.json`.
4. **(F5)** Add `dist`-freshness check to `ci.yml` as a matrix job, or add a `prebuild` script to the server's `package.json`.
5. **(F6)** Add a source-level test that imports via the relative path.
6. **(F7)** Add the default-mode test (or drop it if F2 is adopted).
7. **(F4)** Optional: centralize the `Object.fromEntries(...) as Record<...>` cast.

### Approval recommendation
**Request changes.** The refactor's mechanics are sound but the architectural inversion (server depending on a published client plugin's subpath), the default-mode footgun, and the missing version bump must be resolved before this becomes published surface area. The right call after fixing these is **approve**.

SweetSophia:
Addressed the valid review findings in amended head `206e3cc`.

What changed:
- F1: moved the shared implementation into a neutral `@sweetsophia/noosphere-injected-memory` package. The server now depends on that package, and the OpenClaw plugin depends on/bundles that package. The server no longer depends on an OpenClaw plugin subpath.
- F2/F7: removed the default strip mode. `stripInjectedMemoryBlocks(content, mode)` now requires an explicit mode, and the OpenClaw article-create call passes `OPENCLAW_ARTICLE_CREATE_STRIP_MODE` explicitly.
- F3: removed the new OpenClaw `./injected-memory-strip` public export entirely, so the OpenClaw package no longer adds that public API surface in this PR.
- F4: centralized the `Object.fromEntries(...) as Record<...>` cast in one typed helper inside the neutral package.
- F5: added PR-time package checks for `noosphere-injected-memory` and Docker dist freshness verification for that package.
- F6: kept package-path coverage and added a source-entry agreement test so stale dist/source drift is easier to spot locally.

Verification:
- Focused tests: 25/25 passed.
- `npm run test:memory`: 193/193 passed.
- `npm run typecheck`: passed.
- `npm run lint`: passed with the two existing warnings.
- `DATABASE_URL=postgresql://placeholder@localhost/placeholder npm run build`: passed.
- `noosphere-injected-memory`: typecheck/build/pack passed.
- `openclaw-noosphere-memory`: typecheck/build/pack passed; real tarball inspection confirmed the neutral package is bundled under `node_modules/@sweetsophia/noosphere-injected-memory`.

PR #210, opened 2026-06-23
fix(api): strip injected memory on article writes

### LINKED ISSUE #208: Strip injected memory blocks in POST /api/articles server path
## Context

Review of PR #206 found that `POST /api/articles` currently calls `detectSecretInInputs`, but does not call `stripInjectedMemoryBlocks` before persisting article content.

The OpenClaw package article-create tool strips injected memory blocks client-side before calling the API, but callers that bypass that package can submit content containing blocks such as:

```html
<recall>...</recall>
<hindsight_memories>...</hindsight_memories>
<noosphere_auto_recall>...</noosphere_auto_recall>
```

## Risk

The plugin-side strip is useful defense-in-depth for the OpenClaw path, but it is not a canonical server-side guard. Raw HTTP clients, custom integrations, or future agent frameworks can bypass it.

## Desired follow-up

- Decide whether injected-memory stripping should be enforced server-side in `POST /api/articles`.
- If yes, reuse the canonical strip helper once issue #207 (or its successor) establishes the single source of truth.
- Add regression tests proving direct API article creation cannot persist injected recall/hindsight/noosphere-auto-recall blocks unintentionally.

## Notes

Tracked separately from PR #206 because that PR is a narrow performance/refactor change for the memory-save strip helper.

SweetSophia:
Closes #208.

## Summary

Hardens direct article create/update API writes so injected-memory blocks cannot be persisted by bypassing the agent adapter layer.

## What Changed

- Added shared article API sanitization in `src/lib/api/article-content.ts`.
- `POST /api/articles` now strips injected-memory blocks from `content` and caller-supplied `excerpt` before secret detection, derived excerpt fallback, article persistence, and initial revision creation.
- `PATCH /api/articles/[id]` now strips injected-memory blocks from updated `content` and explicit `excerpt` before persistence and revision creation.
- `PATCH /api/articles/[id]` now runs secret detection on sanitized title/content/excerpt/tags input, closing the update-path bypass found in review.
- Content that becomes empty after stripping is rejected with HTTP 400 instead of creating or updating durable articles with injected-only text.
- Added focused route-level tests for POST and PATCH, including all supported injected block tags, malformed tails, caller-supplied excerpts, injected-only rejection, revision persistence, and visible-secret rejection after stripping.

## Review Follow-Up

The severe review findings were valid for two blocking cases and are addressed here:

- PATCH could previously persist injected-memory blocks and raw revision content.
- POST could previously persist caller-supplied `excerpt` containing injected-memory blocks.

The broader findings for `POST /api/ingest`, `POST /api/answer`, `POST /api/import`, and stripping observability are valid but wider than #208's direct `/api/articles` scope. I opened follow-up issue #211 for those remaining write paths.

## Verification

- Focused route tests: 6/6 passed
  - `articles-post-injected-memory.test.ts`
  - `articles-patch-injected-memory.test.ts`
- `npm run typecheck`: passed
- `npm run lint`: passed with the same 2 pre-existing unused-variable warnings
- `npm run build`: passed
- `npm run test:api`: 329/331 passed
  - Existing unrelated `sync-conflict-review.test.ts` Prisma mock failure reproduced in isolation
  - `bounded-json-routes.integration.test.ts` failed only in the full shared-DB suite and passed in isolation

## Summary

This pull request implements server-side sanitization to strip injected memory blocks (`<recall>`, `<hindsight_memories>`, `<noosphere_auto_recall>`, and malformed tails) from article content and excerpts before they are persisted, addressing issue #208. Previously, this stripping only happened in agent adapters; now it also applies to direct HTTP callers of the article API.

### New utility: `src/lib/api/article-content.ts`
Introduces two helpers:
- **`sanitizeArticleContent`** – Strips injected memory blocks from article body content. Returns a 400 error result (`"Content must include durable text outside injected memory blocks"`) when no durable text remains after stripping.
- **`sanitizeArticleExcerpt`** – Strips injected memory blocks from excerpts, returning an empty string if only injected content was present.

Both reuse the existing `stripInjectedMemoryBlocks` helper from `@sweetsophia/noosphere-injected-memory` with the server-save strip mode.

### Modified: `POST /api/articles` (`src/app/api/articles/route.ts`)
- Sanitizes incoming `content` and `excerpt` before secret detection, persistence, initial revision creation, and excerpt derivation.
- Rejects requests whose content consists entirely of injected memory blocks (HTTP 400).

### Modified: `PATCH /api/articles/[id]` (`src/app/api/articles/[id]/route.ts`)
- Sanitizes incoming `content` and `excerpt` on partial updates before storing them, running secret detection, auto-deriving excerpts from the cleaned text, and creating revisions from the stripped content.
- Rejects content-only-injected-memory payloads and runs secret detection against sanitized fields.
- Revision creation now keys off the sanitized content, so revision history matches what's actually persisted.

### New tests
- **`src/__tests__/api/articles-post-injected-memory.test.ts`** – Verifies POST strips `<recall>`, `<hindsight_memories>`, and `<noosphere_auto_recall>` blocks (and the secrets they contain) from both content and excerpts; truncates malformed (unterminated) injected memory tails; persists stripped content in the initial revision; and rejects content that is entirely injected memory.
- **`src/__tests__/api/articles-patch-injected-memory.test.ts`** – Verifies PATCH applies the same stripping, matches the stripped content in newly created revisions, rejects injected-only updates, and rejects visible secrets that survive stripping.

### DIFF
--- a/src/__tests__/api/articles-patch-injected-memory.test.ts
+++ b/src/__tests__/api/articles-patch-injected-memory.test.ts
@@ -1,0 +1,205 @@
+import assert from "node:assert/strict";
+import crypto from "node:crypto";
+import test from "node:test";
+import type { NextRequest } from "next/server";
+import type { PrismaClient } from "@prisma/client";
+
+if (!process.env.DATABASE_URL) {
+  throw new Error("DATABASE_URL environment variable must be set for tests");
+}
+
+const TEST_RUN_ID = crypto.randomUUID();
+const TEST_PREFIX = `test-issue208-patch-${TEST_RUN_ID}`;
+const TEST_TOPIC_SLUG = `${TEST_PREFIX}-topic`;
+const TEST_KEY_NAME = `${TEST_PREFIX}-key`;
+const TEST_CLIENT_IP = `10.209.${Math.floor(Math.random() * 254) + 1}.${Math.floor(Math.random() * 254) + 1}`;
+
+function buildPatchRequest(
+  articleId: string,
+  body: unknown,
+  rawKey: string,
+): NextRequest {
+  const request = new Request(`http://localhost/api/articles/${articleId}`, {
+    method: "PATCH",
+    headers: {
+      Authorization: `Bearer ${rawKey}`,
+      "Content-Type": "application/json",
+      "x-real-ip": TEST_CLIENT_IP,
+    },
+    body: JSON.stringify(body),
+  });
+  return request as unknown as NextRequest;
+}
+
+async function createWriteKey(
+  prisma: PrismaClient,
+  rawKey: string,
+): Promise<void> {
+  await prisma.apiKey.create({
+    data: {
+      name: TEST_KEY_NAME,
+      keyHash: crypto.createHash("sha256").update(rawKey).digest("hex"),
+      keyPrefix: rawKey.slice(0, 8),
+      permissions: "WRITE",
+      allowedScopes: ["*"],
+    },
+  });
+}
+
+async function setupTopic(prisma: PrismaClient): Promise<{ id: string }> {
+  return prisma.topic.create({
+    data: { name: "Test Issue 208 Patch", slug: TEST_TOPIC_SLUG },
+    select: { id: true },
+  });
+}
+
+async function createArticle(
+  prisma: PrismaClient,
+  topicId: string,
+  slug: string,
+  content = "Original durable content.",
+) {
+  return prisma.article.create({
+    data: {
+      title: `Patch ${slug}`,
+      slug,
+      content,
+      excerpt: "Original excerpt.",
+      topicId,
+      authorName: TEST_KEY_NAME,
+    },
+  });
+}
+
+async function cleanupFixtures(prisma: PrismaClient): Promise<void> {
+  await prisma.article.deleteMany({
+    where: { topic: { slug: TEST_TOPIC_SLUG } },
+  });
+  await prisma.topic.deleteMany({ where: { slug: TEST_TOPIC_SLUG } });
+  await prisma.apiKey.deleteMany({ where: { name: TEST_KEY_NAME } });
+}
+
+test("PATCH /api/articles/[id] strips injected memory blocks before persistence", async () => {
+  const { prisma } = await import("@/lib/prisma");
+  const { PATCH } = await import("@/app/api/articles/[id]/route");
+  const rawKey = `noo_${crypto.randomBytes(32).toString("base64url")}`;
+
+  await cleanupFixtures(prisma);
+  const topic = await setupTopic(prisma);
+  await createWriteKey(prisma, rawKey);
+  const article = await createArticle(prisma, topic.id, "patch-strip");
+
+  try {
+    const content = [
+      "Visible preamble.",
+      "<recall>hidden recall sk-abcdefghijklmnopqrstuvwxyz</recall>",
+      "Visible middle.",
+      "<hindsight_memories>hidden hindsight</hindsight_memories>",
+      "<noosphere_auto_recall>hidden auto</noosphere_auto_recall>",
+      "Visible ending.",
+    ].join("\n");
+
+    const response = await PATCH(
+      buildPatchRequest(article.id, {
+        content,
+        excerpt: "Visible patch summary. <recall>hidden excerpt</recall>",
+      }, rawKey),
+      { params: Promise.resolve({ id: article.id }) },
+    );
+    const body = (await response.json()) as { error?: string };
+
+    assert.equal(response.status, 200, body.error);
+
+    const after = await prisma.article.findUnique({
+      where: { id: article.id },
+      include: { revisions: true },
+    });
+    assert.ok(after, "article must still exist");
+    assert.match(after.content, /Visible preamble/);
+    assert.match(after.content, /Visible middle/);
+    assert.match(after.content, /Visible ending/);
+    assert.doesNotMatch(after.content, /<recall|<hindsight_memories|<noosphere_auto_recall/);
+    assert.doesNotMatch(after.content, /hidden recall|hidden hindsight|hidden auto|sk-abcdefghijklmnopqrstuvwxyz/);
+    assert.match(after.excerpt ?? "", /Visible patch summary/);
+    assert.doesNotMatch(after.excerpt ?? "", /<recall|hidden excerpt/);
+    assert.equal(after.revisions.length, 1, "content PATCH should create one revision");
+    assert.equal(
+      after.revisions[0].content,
+      after.content,
+      "revision content must match stripped article content",
+    );
+  } finally {
+    await cleanupFixtures(prisma);
+  }
+});
+
+test("PATCH /api/articles/[id] rejects content made only of injected memory blocks", async () => {
+  const { prisma } = await import("@/lib/prisma");
+  const { PATCH } = await import("@/app/api/articles/[id]/route");
+  const rawKey = `noo_${crypto.randomBytes(32).toString("base64url")}`;
+
+  await cleanupFixtures(prisma);
+  const topic = await setupTopic(prisma);
+  await createWriteKey(prisma, rawKey);
+  const article = await createArticle(prisma, topic.id, "patch-empty");
+
+  try {
+    const response = await PATCH(
+      buildPatchRequest(article.id, {
+        content: "<recall>hidden only</recall>",
+      }, rawKey),
+      { params: Promise.resolve({ id: article.id }) },
+    );
+    const body = (await response.json()) as { error?: string };
+
+    assert.equal(response.status, 400);
+    assert.equal(
+      body.error,
+      "Content must include durable text outside injected memory blocks",
+    );
+
+    const after = await prisma.article.findUnique({
+      where: { id: article.id },
+      include: { revisions: true },
+    });
+    assert.ok(after, "article must still exist");
+    assert.equal(after.content, article.content);
+    assert.equal(after.revisions.length, 0, "rejected content must not create a revision");
+  } finally {
+    await cleanupFixtures(prisma);
+  }
+});
+
+test("PATCH /api/articles/[id] rejects visible secrets after stripping", async () => {
+  const { prisma } = await import("@/lib/prisma");
+  const { PATCH } = await import("@/app/api/articles/[id]/route");
+  const rawKey = `noo_${crypto.randomBytes(32).toString("base64url")}`;
+
+  await cleanupFixtures(prisma);
+  const topic = await setupTopic(prisma);
+  await createWriteKey(prisma, rawKey);
+  const article = await createArticle(prisma, topic.id, "patch-secret");
+
+  try {
+    const response = await PATCH(
+      buildPatchRequest(article.id, {
+        content: "Visible content with api_key=abcdefghijklmnop",
+      }, rawKey),
+      { params: Promise.resolve({ id: article.id }) },
+    );
+    const body = (await response.json()) as { error?: string };
+
+    assert.equal(response.status, 400);
+    assert.match(body.error ?? "", /secret/);
+
+    const after = await prisma.article.findUnique({
+      where: { id: article.id },
+      include: { revisions: true },
+    });
+    assert.ok(after, "article must still exist");
+    assert.equal(after.content, article.content);
+    assert.equal(after.revisions.length, 0, "secret rejection must not create a revision");
+  } finally {
+    await cleanupFixtures(prisma);
+  }
+});
--- a/src/__tests__/api/articles-post-injected-memory.test.ts
+++ b/src/__tests__/api/articles-post-injected-memory.test.ts
@@ -1,0 +1,186 @@
+import assert from "node:assert/strict";
+import crypto from "node:crypto";
+import test from "node:test";
+import type { NextRequest } from "next/server";
+import type { PrismaClient } from "@prisma/client";
+
+// Route-level test for issue #208. It exercises the real POST /api/articles
+// handler so direct HTTP callers get the same injected-memory protection as
+// agent adapters.
+if (!process.env.DATABASE_URL) {
+  throw new Error("DATABASE_URL environment variable must be set for tests");
+}
+
+const TEST_RUN_ID = crypto.randomUUID();
+const TEST_PREFIX = `test-issue208-${TEST_RUN_ID}`;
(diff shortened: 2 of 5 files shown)


SweetSophia:
## PR #210 Review — strip injected memory on article create (commit `b131a6d`)

### Verdict
**Request changes.** The fix for `POST /api/articles` is correct and well-tested, and the architectural concern from my PR #209 review (neutral package extraction) was properly addressed. But the threat model from issue #208 ("agents can't sneak injected memory blocks into durable storage") is **not contained by this PR** — the PATCH route, the ingest batch endpoint, the answer-to-page endpoint, and the markdown import endpoint all still accept user-supplied content with zero stripping and zero secret detection. Shipping as-is leaves a documented vulnerability that is *wider* than the one being closed.

### Semantic anchors invoked
**Defense in Depth** · **Fail-Safe Defaults** · **DRY / SSOT** · **Hyrum's Law** · **Postel's Law** · **Principle of Least Astonishment** · **MECE** · **STRIDE (Tampering + Information Disclosure)** · **KISS / YAGNI** · **Convention over Configuration**

---

### 🔴 The headline finding: four other write endpoints are still vulnerable

Verified by reading each file in full and grepping for the protected-helper calls:

| Route | File | Has strip? | Has `detectSecretInInputs`? |
|---|---|---|---|
| `POST /api/articles` | `route.ts:303-312` | ✅ (this PR) | ✅ |
| `PATCH /api/articles/[id]` | `[id]/route.ts:139-153, 248` | ❌ | ❌ |
| `POST /api/ingest` | `ingest/route.ts` (286 lines) | ❌ | ❌ |
| `POST /api/answer` | `answer/route.ts` (207 lines) | ❌ | ❌ |
| `POST /api/import` | `import/route.ts` (515 lines) | ❌ | ❌ |
| `POST /api/memory/save` | `memory/save/route.ts:31` | ✅ (via `executeMemorySaveRequest`) | ✅ |

The PR's title and body say "fixes injected-memory smuggling on the public API." That's true for **one of five write endpoints**.

**The PATCH route is the most obvious smuggling route:**
- Line 38: `requirePermission(request, [Permissions.WRITE])` — any WRITE key qualifies.
- Line 147: `updateData.content = content;` — persists raw value, no strip.
- Line 151: `updateData.excerpt = deriveExcerpt(content);` — excerpt derived from un-stripped content.
- Line 248: `content: (content ?? existing.content) as string,` — revision row stores raw value.
- **Bonus:** PATCH has *neither* strip *nor* `detectSecretInInputs` — strictly worse than pre-fix POST.

**Concrete attack (single round-trip):**
1. Authenticate with a WRITE key.
2. `POST /api/articles` with a clean `content` (passes the new strip + existing secret detection).
3. `PATCH /api/articles/:id` with `{"content": "Visible preamble\n<recall>raw secret payload</recall>\n"}`.
4. The raw payload is persisted to `Article.content`, `Article.excerpt` (derived from raw), and `ArticleRevision.content`.

The `ingest` endpoint is even worse — it's a **batch** endpoint, so a single 100-article ingest can poison 100 rows. The `import` endpoint parses `content` out of markdown files inside a zip upload — a high-bandwidth smuggling path.

**Must-fix-in-this-PR vs. follow-up?** PATCH **must** be in this PR — the fix is mechanical (one helper, one call site) and the threat model is explicitly not achieved without it. Ingest/answer/import are appropriate as a follow-up issue with a tracking checklist.

**Apply Defense in Depth + Fail-Safe Defaults:** a single boundary that strips at one entry point is not defense in depth — it's a single point of failure. The correct shape is "strip on every route that calls `prisma.article.create` or `prisma.article.update` with user-controlled content," or, even better, "strip inside a shared helper that all write routes call."

**Recommended shape:**
```ts
// src/lib/api/sanitize-article-content.ts
sanitizeArticleContent(input: { content: string; stripMode: InjectedMemoryStripMode }):
  | { ok: true; content: string; strippedBlocks: string[] }
  | { ok: false; status: 400; error: string }
```
Call from POST, PATCH, ingest, answer, and import. PATCH should also gain `detectSecretInInputs` on the same call (the existing `save.ts:198-206` shows the canonical order: strip → secret-detect). The helper is the SSOT.

---

### 🔴 Caller-supplied `excerpt` is NOT stripped in POST

`src/app/api/articles/route.ts:317` passes `excerpt ?? undefined` to `detectSecretInInputs` (secrets are caught), but the excerpt is NOT passed through `stripInjectedMemoryBlocks`. If a caller supplies `excerpt: "<recall>hidden</recall>"`, the value is stored verbatim. The auto-derived excerpt (line 379: `excerpt || deriveExcerpt(sanitizedContent)`) IS sanitized, so the bypass is only via the caller-supplied path — but that path is the documented one. Both **kilo-code-bot** and **gemini-code-assist** flagged this independently.

---

### 🟡 `strippedBlocks` is silently discarded — observability gap

`route.ts:303-306`:
```ts
const sanitizedContent = stripInjectedMemoryBlocks(content, SERVER_MEMORY_SAVE_STRIP_MODE).content;
```
Only `.content` is read. The `.strippedBlocks` array is thrown away. Compare to:
- `src/lib/memory/api/save.ts:141, 221, 297` — surfaces `strippedBlocks` to the activity log.
- `openclaw-noosphere-memory/src/tools/article-create.ts:107, 171` — surfaces `strippedBlocks` in the response payload.

Two strippers, two observability stories. A future "we suspect someone is sneaking injected memory" investigation has nothing to grep on the article-create path. Trivial two-line fix; should be in this PR.

---

### 🟡 No shared sanitization helper — duplication will drift

The strip + durable-text check + secret-detect pipeline is now implemented in **two places** with subtly different semantics:
- `src/lib/memory/api/save.ts:193-206` — also enforces `minDurableContentLength`, rejects transient-only content.
- `src/app/api/articles/route.ts:303-325` — just requires non-empty trim.

The openclaw plugin at `openclaw-noosphere-memory/src/tools/article-create.ts:131-137` is a *third* re-implementation — same logic, different code, no shared test surface. Extract `sanitizeArticleContent` and `validateMemoryCandidate` helpers; call from all three. (DRY/SSOT.)

---

### 🟢 POST fix itself is correct

Verified the ordering in `route.ts`:
1. Field-level type checks (244-263)
2. Content size check (265-271) — runs on **raw** content; this is correct because the strip is loss-only, so post-strip is always ≤ pre-strip bytes; a 1.5MB all-`<recall>` payload is rightly rejected at the cheap check.
3. **Strip + durable-content check (303-312)** — new, correct.
4. Secret detection (314-325) — runs on **sanitized** content. Correct: a `<recall>sk-…</recall>` is removed before the secret regex sees it.
5. Persistence to `Article.content`, `Article.excerpt` (auto-derived only, line 379), `ArticleRevision.content`. Correct.

---

### 🟢 Architectural concern from PR #209 was addressed

The import is from `@sweetsophia/noosphere-injected-memory` (neutral), not `@sweetsophia/openclaw-noosphere-memory` (client plugin). Verified the package exists at `noosphere-injected-memory/` with its own `package.json`, `src/`, `dist/`. Both the server and the openclaw plugin depend on it via `file:` links. This is exactly the architecture I recommended. ✓

---

### Test quality assessment (MECE)

**What IS tested (3 cases, all POST):**
- ✅ All three block tags stripped; durable text preserved; excerpt derived from stripped; initial revision mirrors stripped.
- ✅ Malformed tail (`<recall>hidden tail` with no close) truncates correctly.
- ✅ All-injected content rejected with the exact 400 message; no row created.

**What is NOT tested (gaps in order of severity):**
- ❌ **PATCH route**: no tests for the unprotected path. Mandatory follow-up test file.
- ❌ **Caller-supplied `excerpt`**: bypass confirmed but not pinned.
- ❌ **Ingest / answer / import**: zero coverage.
- ❌ **strippedBlocks observability**: no test asserts metadata is logged.
- 🟡 **Hardcoded topic slug** (`test-issue208-topic`, line 15) — parallel CI runs race on `prisma.topic.upsert` and `deleteMany`. Pre-existing pattern (matches `articles-patch-restricted-tags.test.ts`), not a regression. Fix opportunistically.

Overall: the tests are tight for what they cover, but they cover ~10% of the surface area this PR is conceptually responsible for.

---

### Behavioral change assessment

**SemVer**: Strictly, this is a backward-incompatible change. A caller posting `{"content": "Some preamble <recall>agent thoughts</recall> postamble."}` to `POST /api/articles` will see the `<recall>…</recall>` silently disappear. The `<recall>` syntax is **not** part of the documented API contract (per `AGENTS.md`), so the change is a bug fix. Document in CHANGELOG under "Security" with a one-liner. No version bump beyond patch-level.

**Postel's Law**: The new behaviour is stricter on accept — it actively rewrites the body to remove a syntactic class. This is the **correct** application of Postel's Law in a security context: the liberal-accept rule is for interoperable protocol data, not for content that flows into a memory layer. The OpenClaw plugin already enforces this; the direct HTTP endpoint is just catching up to the same contract.

**Mode mismatch (minor):** The openclaw plugin uses `OPENCLAW_ARTICLE_CREATE_STRIP_MODE` (throw-on-unclosed); the new server-side strip uses `SERVER_MEMORY_SAVE_STRIP_MODE` (truncate-tail). A content payload that the openclaw client rejects client-side could now succeed server-side with the tail silently dropped. The mismatch is benign (server is safer) but worth documenting.

---

### What CI caught vs. missed

**Caught (all green):**
- `lint-and-typecheck` (40s) — no regressions
- `CodeQL`, `Snyk`, `Analyze` jobs, `docker` build
- 3 in-file tests pass against the real route handler

**Bots that did their job:**
- **kilo-code-bot** identified the PATCH bypass as a 🚨 critical out-of-scope finding, the excerpt bypass as a ⚠️ warning, the `strippedBlocks` observability gap, and the import-ordering nit. Bot's summary: "the whole point of #208 is 'agents can't sneak injected memory blocks into durable storage' — and PATCH is the more obvious smuggling route than POST."
- **gemini-code-assist** identified the caller-supplied excerpt bypass and the parallel-run test risk.
- **sourcery-ai** identified the test setup duplication and the over-aggressive `DATABASE_URL` check.

**Missed (cross-route analysis required):**
- The three other unprotected write endpoints (ingest, answer, import).
- The PATCH route's complete absence of `detectSecretInInputs` (the bots noted the *new* strip gap, not the *predating* secret-detection gap).
- The behavioural mismatch between `OPENCLAW_ARTICLE_CREATE_STRIP_MODE` and `SERVER_MEMORY_SAVE_STRIP_MODE`.

CI signal is good for "the diff is correct and safe in isolation." CI signal is **silent** on "the diff is the right scope." That's the gap the maintainer has to close.

---

1. **(must, in this PR)** Add `stripInjectedMemoryBlocks` + empty-after-strip check to `src/app/api/articles/[id]/route.ts:139-153`. Persist `sanitizedContent` to `Article.content`, `Article.excerpt`, and `ArticleRevision.content`. Add `detectSecretInInputs` for the same field set POST scans. Add a `articles-patch-injected-memory.test.ts` regression test mirroring the existing POST tests. **This is the headline fix.**

2. **(must, in this PR)** Strip caller-supplied `excerpt` in POST (line 317-379). One-line: if `excerpt` is provided, run it through `stripInjectedMemoryBlocks` before the secret scan and before persistence. Add a 1-line regression test.

3. **(should, in this PR)** Surface `strippedBlocks` to the activity log, mirroring `src/lib/memory/api/save.ts:141, 221, 297` and `openclaw-noosphere-memory/src/tools/article-create.ts:107, 171`. Two-line change in the route plus a small assertion in the existing test.

4. **(should, as a follow-up issue)** Extract `src/lib/api/sanitize-article-content.ts` with the shared strip + durable-text + secret-detect pipeline. File a tracking issue with a checklist of the five call sites (POST, PATCH, ingest, answer, import). Reference #208 and this review.

5. **(should, as a follow-up issue)** Add the strip + secret detection to `POST /api/ingest`, `POST /api/answer`, and `POST /api/import`. `import` is the highest-bandwidth smuggling path (zip upload, batch size); prioritise it.

6. **(nit, optional)** Reorder the new third-party import to the top of the import block in `src/app/api/articles/route.ts`.

7. **(nit, optional)** Switch the test topic slug to a per-run UUID and consolidate setup/teardown into `before/after` hooks.

---

### Errors you made — direct assessment

You asked "let us see if I made errors." Direct answers:

1. **You fixed the wrong scope.** Issue #208 says "agents can't sneak injected memory blocks into durable storage." You fixed the POST door. PATCH, ingest, answer, and import are all still wide open. The PATCH route is **strictly worse** than pre-fix POST because it has neither strip nor secret detection. This is the biggest error and it cascades into F1 (no shared helper) and F2 (no caller-supplied excerpt strip).

2. **You silently discarded `strippedBlocks`.** Both other consumers of the shared helper (memory-save, openclaw plugin) surface this for observability. The article-create path throws it away. A two-line change to log it.

3. **The excerpt gap is real.** Caller-supplied excerpt goes to `detectSecretInInputs` but not to `stripInjectedMemoryBlocks`. Inconsistent with how `content` is treated. Both kilo and gemini flagged it; you didn't address it.

The good news: each fix is mechanical. The POST fix itself is well-designed and well-tested. The architectural concern from the prior review (neutral package) was properly addressed. Once the scope is expanded and the helper extracted, this is a clean security PR.

SweetSophia:
Addressed the blocking review findings in amended head `31374d4`.

What changed after review:

- Hardened `PATCH /api/articles/[id]` so updated `content` is stripped before persistence, excerpt derivation, content-change comparison, and revision creation.
- Hardened `PATCH /api/articles/[id]` explicit `excerpt` handling so injected-memory blocks cannot be stored through excerpt updates.
- Added `detectSecretInInputs` coverage to PATCH for sanitized title/content/excerpt/tags input.
- Hardened `POST /api/articles` caller-supplied `excerpt`, which previously bypassed stripping.
- Added shared `src/lib/api/article-content.ts` so POST/PATCH use one article API sanitization path.
- Added focused POST/PATCH route tests covering all injected block tags, malformed tails, injected-only rejection, explicit excerpt stripping, sanitized revision persistence, and visible-secret rejection after stripping.

Verification:

- Focused route tests: 6/6 passed
- `npm run typecheck`: passed
- `npm run lint`: passed with the same 2 existing warnings
- `npm run build`: passed
- `npm run test:api`: 329/331 passed; the two failures are not from this change:
  - existing `sync-conflict-review.test.ts` Prisma mock issue, reproduced in isolation
  - `bounded-json-routes.integration.test.ts` shared-suite DB/order issue, passed in isolation

The broader review points for `POST /api/ingest`, `POST /api/answer`, `POST /api/import`, and stripped-block observability are valid follow-up work; I opened #211 to track them separately.

PR #220, opened 2026-06-25
refine wiki design system and app shell

SweetSophia:
## Summary

This PR refreshes the Noosphere wiki design system and app shell while hardening Redis-backed API rate limiting.

### Design system & branding
- Rename and regroup all global CSS variables into a semantic `--color-*` palette (bg, surface, text, border, accent, status) with backwards-compatible aliases, and introduce shared shadow tiers (`--shadow-sm/md/lg`), spacing (`--space-8`), container widths (`--container-sm/md/lg/xl`), and transition timing tokens.
- Retune light/dark themes around a teal accent (replacing the previous indigo), soften the page background gradients, and add color-mix based shadows for the logo mark and accents.
- Switch app typography from Manrope/JetBrains Mono to **Geist** and **Geist Mono**, and update root + OpenGraph/Twitter metadata descriptions to “Agent-authored documentation, designed for human browsing and editing.”

### Wiki shell & navigation
- Rebuild `src/app/wiki/layout.tsx` into a single responsive `wiki-header-row` grid containing brand, search, nav, and user/auth actions — replacing the prior two-row header.
- Add **Search** as a first-class nav item, rename the settings link to **Recall Settings**, and surface **Activity Log** as a dedicated admin nav item (restoring its discoverability).
- Update `src/app/wiki/wiki.css` to support the new shell, including `wiki-shell`/`page-shell` utilities, redesigned mobile breakpoints, and refined admin nav active/hover styling (using `--accent-color`/`--accent-soft`).
- Remove the redundant radial pseudo-element backdrop and unused `.btn-danger` rules; retune nav link spacing, transitions, and headings (`letter-spacing` normalized to `0` for display fonts).

### Shared UI primitives
- Introduce `.section-card`, `.stack-sm/md/lg`, `.cluster`, `.meta-row`, and `.badge-row` layout utilities for wiki/admin pages.

### API rate limiting & readiness
- Switch `src/lib/rate-limit.ts` to `getReadyRedisClient()` so the limiter awaits lazy Redis initialization before issuing commands.
- Extend `src/__tests__/rate-limit.test.ts` with corrected `-inf` score handling, async pipeline `exec`, and explicit allowance assertions across the full request window.
- Add a comprehensive `FakeRedisClient` to `src/__tests__/security/proxy.test.ts` (with sorted-set + pipeline semantics, `-inf` handling, and a readiness guard) and wire `_redisTestHooks.setClientForTesting`/`reset` into `beforeEach`/`afterEach` so proxy tests exercise Redis-backed rate limiting paths.

---

This pull request refines the wiki's design system and application shell by extracting a shared testing helper and adding new test coverage for the rate limiter's lazy Redis connection behavior.

**Key Changes:**

1. **New shared test helper (`src/__tests__/_helpers/fake-redis.ts`)**: Extracted the `FakeRedisClient` implementation into a reusable helper module. This version supports configurable connection delays and the ability to simulate connection failures, giving tests more control over the client lifecycle.

2. **Test refactoring**: Both `rate-limit.test.ts` and `proxy.test.ts` were updated to import the shared `FakeRedisClient` instead of defining their own inline copies, removing duplication.

3. **New rate limit tests**: Added two edge-case tests that verify:
   - When the Redis client is still connecting, a burst of concurrent `rateLimit` calls shares the single in-flight connection attempt rather than each initiating their own (verifying only one `connect` call occurs).
   - When the lazy Redis connect fails, the rate limiter falls open (allows the request) and suppresses immediate retries through a reconnect cooldown, avoiding repeated connection attempts and log noise on subsequent requests within the cooldown window.


SweetSophia:
## Code Review — refine wiki design system and app shell

Reviewed all 8 files. Solid, well-scoped PR. The rate-limit hardening is a genuine correctness fix, the design-token refactor is a non-breaking migration, and the admin-nav discoverability work is a clear win. No blockers — a few nits and one maintainability item below. Terminology uses semantic anchors where useful.

---

### ✅ What's genuinely good

**1. `getReadyRedisClient()` is the right fix (graceful degradation / fail-open).**
`src/lib/rate-limit.ts:29` previously called `getRedisClient()`, which returns the singleton even while its status is `"wait"` (lazy connect). Because the client is created with `enableOfflineQueue: false`, enqueuing a pipeline against a non-ready client would error, get caught, and fail open anyway — so the old code *appeared* to work but never actually connected. Awaiting readiness is the correct behavior. The fail-open fallback (`return { allowed: true }`) is a sound availability choice for a rate limiter.

**2. The test-double corrections in `rate-limit.test.ts` are all accurate.**
- `-inf` handling: `parseFloat("-inf")` → `NaN`, which breaks the range filter. `min === "-inf" ? Number.NEGATIVE_INFINITY : parseFloat(min)` matches real Redis semantics.
- Timestamp index: member format is `${now}:${uuid}` (rate-limit.ts:39), so `split(":")[0]` is the timestamp — the old `[1]` parsed the UUID and yielded `NaN`. Correct fix.
- `exec` → `async` and the explicit allowance assertions align with the implementation, which reads `zcard` *before* `zadd`, so the first `maxRequests` calls are allowed and the next is blocked.

**3. Design-token refactor is a non-breaking migration.**
Introducing canonical `--color-*` tokens while retaining the legacy names as aliases is a pragmatic **backwards-compatible shim** — no consumer breaks. `100dvh` over `100vh` is a correct switch to **dynamic viewport units** (handles mobile browser chrome).

**4. `.btn-danger` removal is not a regression.**
`wiki.css` had *two* `.btn-danger` blocks; this PR removes the older outline (`color-mix`) variant and keeps the solid gradient definition at `wiki.css:1363`. The four delete-button components (`DeleteArticleForm`, `DeleteTopicButton`, `DeleteTagButton`, `KeyActionButtons`) remain styled.

---

### 🔎 Suggestions

**1. Duplicate Test Doubles — extract a shared fake (DRY).**
`rate-limit.test.ts` and `security/proxy.test.ts` now each ship their own `FakeRedisClient` with *different* storage models (JSON-stringified `["ts:uuid", …]` vs `Map<string, {score, member}[]>`). Two test suites, two truths about the same dependency. Per **DRY**, lift one into `__tests__/helpers/` so a future Redis-API change can't make the fakes diverge silently. (Independent bots flagged this too.)

**2. Alias sprawl dilutes the Single Source of Truth (SSOT).**
`globals.css` now carries two complete naming schemes side by side (`--color-*` canonical + `--surface-color`/`--text-color`/`--accent`/… aliases). This is fine as a migration step, but without a plan to move consumers to the canonical names and drop the aliases, drift sets in and both lists must be edited forever. Consider a follow-up to migrate + retire the legacy names (and note any light/dark asymmetries while you're in there).

**3. Fail-open is correct, but make it observable.**
When Redis is down, rate limiting silently becomes a no-op (`allowed: true`). That's the right availability tradeoff, but it's invisible. A metric or a single warn-rate-limited log on the `return { allowed: true }` path would make an outage diagnosable rather than latent.

**4. Concurrency at first connect — confirm under burst.**
On the first request(s) the singleton is in `"wait"`; concurrent callers can each reach `await client.connect()` before anyone flips to `"ready"`. ioredis tolerates concurrent `connect()`, but worth a quick burst test to rule out duplicate-connect noise.

**5. Pre-existing: rejected requests still occupy a slot.**
The pipeline runs `zadd` unconditionally in the same batch as `zcard`, so a blocked request still gets inserted into the ZSET. This is **not introduced** by this PR — but the new assertions lock the behavior in. If you ever want strict "don't count the denied request" semantics, `zadd` needs to be conditional on the count (e.g. a second round-trip or `ZADD GT`/Lua). Optional.

**6. QA call-out: delete buttons change appearance.**
With the outline variant gone, delete buttons now render as a solid red gradient rather than the previous tinted outline. Intentional per the token cleanup, but it's a user-visible change worth a visual sanity check across the four components.

---

### 📌 Note on `letter-spacing: 0` for display text
Normalizing tracking to `0` on Fraunces is a deliberate taste call. Large display serifs conventionally take slight *negative* tracking, so headings may read a touch loose — fine if that's the intended editorial feel, just flagging it.

---

**Verdict:** Ship it. Address the duplicate test double (extract a shared helper) as a near-term follow-up; the rest are nits.

SweetSophia:
Reviewed Sophie's comment: valid overall, with no blockers.

Accepted and fixed in commit `8b90dcd813ec1d633e8d31cdc6e4ff56831d7f80`:
- Extracted the duplicate Redis test doubles from `rate-limit.test.ts` and `security/proxy.test.ts` into `src/__tests__/_helpers/fake-redis.ts`.
- Added direct lazy-client burst coverage proving first-request concurrency shares one Redis connect attempt.
- Tightened `getReadyRedisClient()` with an in-flight `redisConnectPromise` single-flight guard.
- Added connect-failure coverage proving the rate limiter fails open and the reconnect cooldown suppresses immediate retries.
- Cleared the in-flight connect promise in `_redisTestHooks.setClientForTesting()` for safer test isolation.
- Added comments explaining the timing/log-count assertions.

Classified as follow-up/non-blocking:
- Full token alias migration/removal is a later design-system cleanup, not needed for PR-1 compatibility.
- Fail-open observability would be useful, but changing logging/metrics behavior is outside this shell/foundation PR.
- Rejected requests occupying a ZSET slot is pre-existing rate-limiter semantics, not introduced here.
- Search-cache has its own richer Redis fake; worth consolidating later, but the review's rate-limit/proxy duplicate-fake issue is fixed.

Local verification passed:
- `npm run test:rate-limit` (12/12)
- `npm run test:security` (15/15)
- `npm run test:cache` (8/8)
- `npm run typecheck`
- `npm run lint` (existing `descendants` warning only)
- full `npm run test` with explicit local Docker `DATABASE_URL`
- `npm run build`
- `npm audit --audit-level=high`
- `npm run package-policy:check`
- `git diff --check`

Two independent validation passes found no blockers; valid minor findings were addressed before push. GitHub checks are still running on the new head at the time of this comment.

SweetSophia:
## Final Review — LGTM ✅

Verified the follow-up commit (`8b90dcd test: consolidate redis rate-limit fakes`) against the source and ran the affected suites:

```
test:rate-limit → 12/12 pass (incl. both new burst/fail-open tests)
security/proxy  → 5/5 pass (refactored onto the shared fake)
```

### Previous suggestions — all addressed

**1. Duplicate Test Double → resolved (DRY).**
Both suites now import a single `FakeRedisClient` from `src/__tests__/_helpers/fake-redis.ts`, unified on the `{score, member}` sorted-set storage model. Bonus: `get`/`setex`/`expire` TTL semantics and test-only knobs (`connectDelayMs`, `rejectConnect`) are centralized. One source of truth for Redis behavior now.

**2. First-connect concurrency → resolved (single-flight).**
`redis.ts` now memoizes the connect promise:
```ts
redisConnectPromise ??= client.connect().finally(() => { redisConnectPromise = null; });
await redisConnectPromise;
```
This is a correct **single-flight / memoized-promise** pattern: concurrent callers in the `"wait"` window share one in-flight `connect()`, so `connect()` is invoked at most once per burst — eliminating any duplicate-connect hazard. The `.finally()` clears the memo so a later re-initiation (post-reset) still works, and the memo is correctly reset in `resetRedisClient`, `closeRedisClient`, and `setClientForTesting`. The new test `"shares the first Redis connection attempt across a lazy-client burst"` proves `connectCalls === 1` across 8 concurrent requests. 👍

**3. Fail-open observability → resolved.**
The new `"falls open and suppresses immediate retries when the lazy Redis connect fails"` test pins the cooldown path: a failed connect logs exactly once, the reconnect cooldown short-circuits the next call, and the limiter fails open. The outage behavior is now asserted rather than implicit.

### Remaining (all nits — non-blocking)

- **SSOT alias retirement** — the canonical `--color-*` tokens + legacy aliases still coexist; fine for now, worth a tracked follow-up to migrate consumers and drop the legacy names so both lists don't drift.
- **Pre-existing** — rejected requests still occupy a ZSET slot (`zadd` runs unconditionally in the pipeline); not introduced here, now locked in by tests. Optional to revisit if strict denied-request accounting is ever wanted.
- **QA** — delete buttons render as solid gradient (outline variant removed); intentional, just flagging for a visual check.

### Verdict
Ship it. The rate-limit hardening is correct, the test consolidation is clean, and the single-flight connect is a nice piece of concurrency hardening on top of the original fix. 🚀

PR #221, opened 2026-06-25
refine wiki home page

SweetSophia:
## Summary

This pull request refines the wiki home page layout and information hierarchy to improve scannability and clarity around key metrics and content sections.

### Header Metrics Reorganized
- The header pill metrics now highlight **total articles**, **mapped topics**, and **topic clusters** instead of the previous "recent updates surfaced" count
- Added a dedicated `wiki-home-hero` class with bottom border separator and adjusted page description copy
- Replaced inline admin button styles with a reusable `cluster` class

### Featured Latest Article with Secondary Updates
- Split the "Recently updated" section into a **featured latest article** (displayed prominently with more tags and larger typography) and a **compact secondary list** for additional updates
- Introduced a shared `RestrictedArticleIcon` component to reduce duplication of the access-restricted lock indicator
- Added locale-aware `homeDateFormatter` for consistent date display across both featured and secondary update cards
- Replaced the old `.recent-updates-grid` / `.recent-updates-list` layout with a new `.home-updates-layout` two-column design

### Topic Tree Improvements
- Added a `directArticleCount` field to `TopicRenderNode` to surface the number of articles directly attached to each topic (separate from total articles including descendants)
- Converted the entire topic card into a clickable link (`<Link>` wrapping the whole `.topic-tree-card`), removing the previous stretched-link pseudo-element technique
- Added a `depth` prop with a `Level {n}` indicator inside the kind label for nested topics
- Introduced visual differentiation for root topics via a `data-depth="0"` selector with a gradient background
- Each topic card now shows both direct article count and nested topic count in a meta row

### Supporting Styles
- Added `.wiki-home-section` spacing utility applied to recent updates and topics sections
- Refactored responsive grid rules so the new `.home-updates-layout` handles stacking on small screens
- Updated media query selector from `.recent-updates-shell` to `.home-updates-layout`

## Verification
- `npm run lint`
- `npm run typecheck`
- `npm run package-policy:check`
- `npm audit --audit-level=high`
- `npm run build` with PostgreSQL connection
- `npm run test` with PostgreSQL connection
- `git diff --check`

## Review Notes
- Two independent review passes were completed
- First pass addressed: nested topic level label, locale-neutral dates, removal of legacy recent-updates CSS
- Second pass found no blockers

---

## Summary

This PR refines the wiki home page to improve accessibility, labeling clarity, and visual consistency.

### Accessibility Improvements
- **Standardized date formatting**: Changed the home page date formatter from locale-default to `en-GB` for consistent display across users.
- **Enhanced screen reader support**: 
  - Added descriptive `aria-label` attributes to topic tree cards and article cards summarizing their key information (article counts, restriction status, topic placement).
  - Marked decorative/duplicate information (kicker labels, dates, level indicators) as `aria-hidden="true"` so screen readers don't repeat the data.
  - Replaced the `RestrictedArticleIcon`'s `title` attribute (tooltip) with `role="img"` and `aria-label="Restricted"` for proper assistive technology support.
- **Removed redundant accessible name**: Removed the standalone `aria-label` on the secondary articles container since each card now has its own descriptive label.

### Content & Copy Refinements
- Renamed the "topic clusters" metric to "branch topics" and added a helper function (`countBranchTopics`) that recursively counts all branch topics throughout the topic tree, not just root-level clusters.
- Replaced ad-hoc pluralization with a shared `pluralize` helper for consistency.
- Updated the topic tree meta to label the count as "total articles" for clarity.
- Removed the redundant nested topics count from the meta row (it was already shown in the metrics block).
- Removed fallback `?? []` on `restrictedTags` since the property is already typed as an array.

### Styling Adjustments
- Replaced the hardcoded `rgba(95, 115, 255, 0.18)` gradient with the existing `var(--accent-glow)` token for design system consistency.
- Removed an unused `.article-card-grid` rule.
- Added a rule so a single child in the updates layout spans the full grid width.
- Removed the `min-height: 100%` from the featured update card and `position: relative` from the topic tree card to let layout flow more naturally.
- Updated the topic tree level indicator color from `--text-faint` to `--text-muted` for better contrast.

---

## Description

Improved accessibility of the wiki home page by enhancing the screen reader labels for restricted articles. The labels now include the specific restriction tags, providing more informative context for assistive technology users when navigating restricted content.

### Changes
- Updated the `RestrictedArticleIcon` component to expose the restriction tags in its `aria-label` (e.g., "Restricted: internal, confidential").
- Enhanced the `articleCardLabel` function to include the restriction tags in the accessible name of article cards (e.g., "Restricted article (internal, confidential): Title in Topic"), so the tags are announced along with the article title and topic.

### DIFF
--- a/src/app/wiki/page.tsx
+++ b/src/app/wiki/page.tsx
@@ -19,6 +19,7 @@ interface ArticleCountMap {
 
 interface TopicRenderNode {
   node: TopicNode;
+  directArticleCount: number;
   articleCount: number;
   descendantCount: number;
   children: TopicRenderNode[];
@@ -33,32 +34,69 @@ type WikiHomeArticle = Prisma.ArticleGetPayload<{
 
 function buildTopicRenderNode(node: TopicNode, countMap: ArticleCountMap): TopicRenderNode {
   const children = node.children.map((child) => buildTopicRenderNode(child, countMap));
+  const directArticleCount = countMap[node.id] ?? 0;
 
   return {
     node,
-    articleCount:
-      (countMap[node.id] ?? 0) + children.reduce((total, child) => total + child.articleCount, 0),
+    directArticleCount,
+    articleCount: directArticleCount + children.reduce((total, child) => total + child.articleCount, 0),
     descendantCount:
       node.children.length + children.reduce((total, child) => total + child.descendantCount, 0),
     children,
   };
 }
 
-function TopicTreeNode({ tree }: { tree: TopicRenderNode }) {
-  const { node, articleCount, descendantCount, children } = tree;
+const homeDateFormatter = new Intl.DateTimeFormat(undefined, {
+  month: "short",
+  day: "numeric",
+  year: "numeric",
+});
+
+function RestrictedArticleIcon({ tags }: { tags: string[] }) {
+  if (tags.length === 0) return null;
+
+  return (
+    <span className="restricted-icon" title={`Restricted: ${tags.join(", ")}`}>
+      <svg
+        width="12"
+        height="12"
+        viewBox="0 0 24 24"
+        fill="none"
+        stroke="currentColor"
+        strokeWidth="2"
+        strokeLinecap="round"
+        strokeLinejoin="round"
+        aria-hidden="true"
+      >
+        <rect x="3" y="11" width="18" height="11" rx="2" ry="2" />
+        <path d="M7 11V7a5 5 0 0 1 10 0v4" />
+      </svg>
+    </span>
+  );
+}
+
+function TopicTreeNode({ tree, depth = 0 }: { tree: TopicRenderNode; depth?: number }) {
+  const { node, directArticleCount, articleCount, descendantCount, children } = tree;
   const hasChildren = children.length > 0;
 
   return (
-    <div className="topic-tree-node">
-      <div className="topic-card topic-tree-card">
+    <div className="topic-tree-node" data-depth={depth}>
+      <Link href={`/wiki/${node.slug}`} className="topic-card topic-tree-card">
         <div className="topic-tree-copy">
-          <p className="topic-tree-kind">{hasChildren ? "Topic cluster" : "Leaf topic"}</p>
+          <p className="topic-tree-kind">
+            {hasChildren ? "Topic cluster" : "Leaf topic"}
+            {depth > 0 ? <span>Level {depth}</span> : null}
+          </p>
           <h2 className="topic-tree-title">
-            <Link href={`/wiki/${node.slug}`}>{node.name}</Link>
+            <span>{node.name}</span>
           </h2>
           <p className="topic-tree-description">
             {node.description ?? "A focused pocket of the wiki ready for articles, references, and linked subtopics."}
           </p>
+          <div className="topic-tree-meta meta-row">
+            <span>{directArticleCount} direct article{directArticleCount !== 1 ? "s" : ""}</span>
+            <span>{descendantCount} nested topic{descendantCount !== 1 ? "s" : ""}</span>
+          </div>
         </div>
 
         <div className="topic-tree-metrics" aria-label={`${node.name} metrics`}>
@@ -71,12 +109,12 @@ function TopicTreeNode({ tree }: { tree: TopicRenderNode }) {
             <span>subtopic{descendantCount !== 1 ? "s" : ""}</span>
           </div>
         </div>
-      </div>
+      </Link>
 
       {hasChildren && (
         <div className="topic-tree-children">
           {children.map((child) => (
-            <TopicTreeNode key={child.node.id} tree={child} />
+            <TopicTreeNode key={child.node.id} tree={child} depth={depth + 1} />
           ))}
         </div>
       )}
@@ -115,13 +153,17 @@ export default async function WikiHomePage() {
 
   const totalArticles = Object.values(countMap).reduce((total, count) => total + count, 0);
   const topicTree = roots.map((topic) => buildTopicRenderNode(topic, countMap));
+  const topicClusters = topicTree.filter((topic) => topic.children.length > 0).length;
+  const latestArticle = recentArticles[0];
+  const secondaryArticles = recentArticles.slice(1);
 
   return (
     <div className="wiki-content wiki-home">
       <PageHeader
         eyebrow="Knowledge atlas"
         title="Noosphere"
-        description="Agent-authored documentation with a calmer information hierarchy: browse live updates, scan topic clusters, and dive straight into the wiki's deepest branches."
+        description="Agent-authored documentation with clear topic paths, readable update trails, and fewer dead ends between question and answer."
+        className="wiki-home-hero"
         meta={
           <div className="page-meta-pills">
             <span className="page-meta-pill">
@@ -133,14 +175,14 @@ export default async function WikiHomePage() {
               <span>topics mapped</span>
             </span>
             <span className="page-meta-pill">
-              <strong>{recentArticles.length}</strong>
-              <span>recent updates surfaced</span>
+              <strong>{topicClusters}</strong>
+              <span>topic clusters</span>
             </span>
           </div>
         }
         actions={
           isAdmin ? (
-            <div style={{ display: "flex", gap: "0.5rem", flexWrap: "wrap" }}>
+            <div className="cluster">
               <Link href="/wiki/admin/topics" className="btn btn-secondary btn-sm">
                 New Topic
               </Link>
@@ -153,12 +195,12 @@ export default async function WikiHomePage() {
       />
 
       {recentArticles.length > 0 && (
-        <section className="browse-section">
+        <section className="browse-section wiki-home-section wiki-home-recent">
           <div className="section-header">
             <div className="section-header-copy">
               <p className="page-eyebrow">Fresh signal</p>
               <h2 className="section-title">Recently updated</h2>
-              <p className="section-subtitle">The newest edits, additions, and refinements across the wiki.</p>
+              <p className="section-subtitle">The newest edits, additions, and refinements are separated from the topic map so the landing page stays easy to scan.</p>
             </div>
             <div className="section-actions">
               <Link href="/wiki/search" className="btn btn-secondary btn-sm">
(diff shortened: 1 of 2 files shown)


SweetSophia:
Addressed the valid review findings in `5dd1478`:

- made the restricted marker accessible with `role="img"` + `aria-label` instead of relying on `title`
- made the featured update span both grid columns when it is the only recent article
- changed topic cluster counting to include nested clusters, not only top-level roots
- removed the duplicate nested-topic line from the topic card metadata; the metric rail remains the single subtopic count
- pinned home date formatting to a stable locale

The mobile breakpoint concern is already covered: `.home-updates-layout` collapses to `grid-template-columns: 1fr` in the existing `@media (max-width: 960px)` block.

Local verification after the patch:
- `npx eslint src/app/wiki/page.tsx`
- `npm run typecheck`
- `git diff --check`
- `DATABASE_URL=... npm run build`

SweetSophia:
## In-Depth Review — PR #221 "refine wiki home page"

**Verdict: Shippable after the items below — but you did make a few errors worth fixing first.** No SSR/hydration bugs; the structural refactor (whole-card link, `directArticleCount`, depth indicators, featured/secondary split, dropped stretched-link hack) is sound. Verified the actual source against `origin/master` (CSS `display` on the new anchor-card, mobile stacking, orphan classes) and dispatched an independent deep pass for a11y / info-architecture. Findings reconciled below.

---

### 🐛 Real bugs / errors introduced by this PR

**B1. `.article-card-grid` is now orphaned CSS (wiki.css:579–582).**
No JSX in `src/` references it after this PR (you replaced it with `.article-card-featured` / `.article-card-compact`). Dead code. **Fix:** delete the rule. *(YAGNI)*

**B2. The "articles" stat label now contradicts the new "X direct articles" meta row (page.tsx ~111).**
The card shows **both**:
- `topic-tree-meta`: "5 direct articles"
- `topic-tree-stat`: "13 articles"

A reader sees two numbers with no clue that the right-hand one includes descendants. The old code had only one number so the label was fine; the split broke the meaning. **Fix:** change the right-hand label to `articles in tree` or `total articles`. (`subtopics` is still fine — it only counts descendants.)

**B3. "topic clusters" hero metric is internally redundant and can read absurd (page.tsx:163, 186).**
`countTopicClusters` returns the count of *every* internal node — the same thing the tree below already displays. With a wiki of 12 mapped topics you'll happily get "85 topic clusters" next to "12 topics mapped", and the kicker inside every card already labels each internal node a "Topic cluster". The number is a numeric re-statement of visible structure. **Fix options (design call):**
- redefine as **non-leaf topics** and rename the pill to match (`branch topics` / `non-leaf topics`), or
- redefine as **roots** (`roots.length` — distinct from `allTopics.length` because some topics have orphan parents), or
- drop the pill; the tree communicates this.

**B4. `restrictedTags ?? []` is dead nullish coalescing (page.tsx:230, 261).**
Schema is `restrictedTags String[] @default([])` — never null. The old code had the same smell with a ternary; carrying it forward is a small lie to the reader. **Fix:** drop the `?? []` in both spots.

**B5. Featured tag slice silently changed `3 → 4` (page.tsx:238).**
Defensible (the featured card is wider) but invisible in the diff. Worth a commit-message note so future reviewers don't miss the intent.

---

### ♿ Accessibility (WCAG)

**A1. Whole-card link names are noisy (page.tsx:59, ~92, ~229, ~260).**
`.topic-tree-card` / `.home-featured-update` / `.home-update-card` are all `<Link>`. Their accessible name is the concatenation of every text descendant: kicker + date + restricted label + title + excerpt + tags. A screen-reader user navigating the home hears, e.g., *"Engineering 14 Mar 2026 Restricted: financial, internal, draft Title. Excerpt… tag1 tag2…"* for every card. **Fix:**
- `aria-hidden="true"` on `.topic-tree-kind` (page.tsx:94) — duplicates the title.
- `aria-hidden="true"` on `.article-kicker` and `.article-date` (page.tsx:226, ~257) — same reason.
- Optionally tighten `RestrictedArticleIcon`'s label to `"Restricted"` (the tags are already announced via the article model).

**A2. `aria-label="Additional recent updates"` on a `<div>` does nothing (page.tsx:249).**
Per ARIA, `aria-label` on an element without a landmark role is ignored by most AT. The wrapping `<section>` already announces "Recently updated" via its heading. **Fix:** delete the attribute.

**A3. `aria-label={`${node.name} metrics`}` on a role-less `<div>` inside the link (page.tsx:109) is noise.**
The metric tiles' text is already part of the link's accessible name; the label neither overrides nor adds a landmark. **Fix:** delete it, or mark the tiles `aria-hidden="true"` and add a single `.sr-only` summary.

**A4. "Level {depth}" span is announced as part of every topic card's link name (page.tsx:96).**
"Level 2 Engineering …" is purely structural decoration. **Fix:** `aria-hidden="true"` on the `<span>`; visual stays.

**A5. `--text-faint` contrast on cream fails WCAG 2.1 SC 1.4.3 for the new "Level N" text.**
~`#81879b` on `#fffaf1` ≈ **3.4:1**, below the 4.5:1 threshold for small text. The new "Level N" inherits this at `0.72rem`. Pre-existing for `.article-date` / stat labels, but this PR surfaces it on a new element. **Fix:** bump this specific span to `var(--text-muted)`, or darken `--color-text-faint`.

**A6. `RestrictedArticleIcon` `role="img"` + `aria-label` pattern — correct ✓.**
SVG is `aria-hidden`, the span carries the label, the component returns `null` when no tags. The WAI-ARIA APG-blessed pattern.

**A7. `<h2>`/`<h3>` inside an `<a>` card-link — correct ✓.**
Valid HTML5 and well-handled by modern screen readers.

---

### 📐 CSS / layout (verified — no issues, a few cleanups)

- `.topic-tree-card` keeps `display: grid` (wiki.css:723), so wrapping the card in `<Link>` doesn't collapse it. ✓
- The 2-col `.home-updates-layout` correctly stacks to `1fr` in the `@media (max-width: 960px)` block (wiki.css:2073–2082). ✓
- `:only-child` span (wiki.css:677) handles the lone-featured-or-lone-list case correctly. ✓
- `data-depth="0"` gradient selector (specificity 0,2,1) wins over the base `.topic-card` background. ✓
- Children of a node render outside the `<a>`, so clicks on child cards don't propagate to the parent. ✓

Cleanups while you're in there:
- `position: relative` on `.topic-tree-card` (line 722) is now vestigial — it existed only for the stretched-link `::after` trick that's gone. *(dead-code)*
- `.home-updates-layout` uses `align-items: stretch` + `.home-featured-update` has both `min-height: 100%` *and* `align-content: start`. Belt-and-suspenders; `align-items: stretch` on the parent is enough. *(KISS)*

---

### 🔧 Nits / maintainability

**M1. `RestrictedArticleIcon` is named "shared" but has one call site (page.tsx:59).**
Either move it to `src/components/wiki/RestrictedArticleIcon.tsx` to honour the name, or rename to e.g. `RecentArticleLockIcon` so the abstraction doesn't promise more than it delivers.

**M2. Featured card gradient is a hardcoded indigo `rgba(95,115,255,…)` (wiki.css:570–572).**
Doesn't match the teal accent (PR 220). You didn't introduce the hardcoded color, but this PR promotes `.article-card-featured` to the home hero, so the inconsistency is now prominently visible. **Fix:** swap to `color-mix(in srgb, var(--accent-color) 18%, transparent)` or equivalent — restore **Single Source of Truth** for accents.

**M3. `TopicRenderNode` materializes three counts (page.tsx:22–24).**
`articleCount` and `descendantCount` are pure functions of the tree + `directArticleCount`. Materializing all three is fine for clarity; could be derived at render time if you ever care. *(Minor.)*

**M4. `homeDateFormatter` (page.tsx:49).**
Server-component module-scope, locale-hardcoded → hydration-safe. ✓ If another page ever needs the same format, lift to `src/lib/`.

---

### ✅ Suggested fix order

1. **B2** (label ambiguity) — one-line copy fix; highest user-facing impact.
2. **B3** (cluster metric) — pick a definition; design call.
3. **B1** (delete `.article-card-grid`).
4. **A2, A3** (delete two ineffective `aria-label`s).
5. **A4** (`aria-hidden` on "Level N").
6. **A1** (hide kicker/date from link name).
7. **B4** (drop `?? []`).
8. **M2** (swap indigo → accent token).
9. A5 + the rest as a separate pass.

**Note on methodology:** I read the actual `wiki.css` and `page.tsx` to verify layout / cascade claims and dispatched an independent deep review for a11y and info-architecture. No emoji-praise for the parts you got right; the structural refactor is genuinely clean — but B2 and the aria-label mistakes are worth catching before this lands.

SweetSophia:
Addressed the valid points from the latest full review in `d7df9b7`:

- removed dead `.article-card-grid` CSS
- changed the topic metric wording to `branch topics` and kept the recursive count
- clarified topic-card metric language to `total articles`
- removed ineffective role-less `aria-label`s
- hid decorative kicker/date/level/metric text from link accessible names and added concise link labels
- dropped stale `restrictedTags ?? []` fallbacks
- switched the featured-card accent wash from hardcoded indigo to `var(--accent-glow)`
- removed vestigial `position: relative` and redundant featured-card `min-height`
- bumped the level-label color to `var(--text-muted)` for better contrast

Not changed:
- `.wiki-home-recent` / `.wiki-home-topics` stay as semantic section hooks; no visual or accessibility impact.
- The featured-card tag slice remains at 4 intentionally because the featured card has more horizontal space than the compact cards.

Local verification before push:
- `npm run lint` (passes with existing `descendants` warning)
- `npm run typecheck`
- `npm run package-policy:check`
- `npm audit --audit-level=high`
- `DATABASE_URL=... npm run build`
- `DATABASE_URL=... npm run test`
- `git diff --check`
- desktop/mobile browser smoke on `/wiki` with no horizontal overflow or page errors

Second independent review pass is running now; I will report any findings if it catches something.

SweetSophia:
Kilo caught one valid follow-up after `d7df9b7`: my restricted-label cleanup went too far and stopped exposing the actual restricted scope names to assistive tech. Fixed in `2e6dcfb`:

- `RestrictedArticleIcon` now announces `Restricted: <scopes>` again.
- Whole-card article labels now include `Restricted article (<scopes>)` when applicable.

Verification after the patch:
- `npx eslint src/app/wiki/page.tsx`
- `npm run typecheck`
- `git diff --check`

SweetSophia:
## Re-Review — PR #221 (after `2e6dcfb`)

**Verdict: LGTM — ship it.** 🎯

Every substantive finding from the first review is resolved and I re-verified against the actual source rather than trusting the diff.

### Verification run
- `npx tsc --noEmit` → **clean** (exit 0)
- `eslint src/app/wiki/page.tsx` → **clean** (exit 0)
- Working tree matches PR tip `2e6dcfb` exactly
- Contrast of `--text-muted` (`#5e6375`) on cream (`#fffaf1`) recomputed: **≈5.74:1** — clears WCAG AA 4.5:1 (old `--text-faint` was ~3.4:1). The A5 fix is real, not a token swap to another failing value.

### Findings checklist

| # | Finding | Status |
|---|---------|--------|
| **B1** | Orphan `.article-card-grid` | ✅ Deleted (wiki.css) |
| **B2** | "articles" vs "direct articles" contradiction | ✅ Right stat now reads **"total articles"**; topic-card `aria-label` also disambiguates |
| **B3** | "topic clusters" redundant/over-counts | ✅ Renamed → **"branch topics"** (`countBranchTopics`) — label now matches the math |
| **B4** | Dead `?? []` on `restrictedTags` | ✅ Removed at both call sites |
| **A1** | Noisy whole-card link names | ✅ Kickers/dates `aria-hidden`, plus explicit `aria-label`s (`articleCardLabel`, topic-card summary) |
| **A2** | Dead `aria-label` on `.home-update-list` | ✅ Removed |
| **A3** | Dead `aria-label` on `.topic-tree-metrics` | ✅ Now `aria-hidden="true"` (link carries the summary instead) |
| **A4** | "Level N" in link name | ✅ `aria-hidden="true"` on the span |
| **A5** | `--text-faint` contrast | ✅ `.topic-tree-kind span` → `--text-muted` (verified 5.74:1) |
| **M2** | Hardcoded indigo gradient | ✅ → `var(--accent-glow)` (SSOT restored) |
| **C2** | Belt-and-suspenders `min-height: 100%` | ✅ Dropped; `align-items: stretch` alone now |
| **C4** | Vestigial `position: relative` | ✅ Removed |

The `pluralize` helper and the two `aria-label` builders read correctly across edge cases (0/1/many, empty tags). The whole-card `<Link>` + nested `<h2>`/`<h3>` is valid and the `aria-label` correctly overrides child text for the link's accessible name — so no double-announcement.

### One truly optional nit (don't block on this)
The "restore restricted scope labels" commit now conveys the scope tags in **two** spots with slightly different phrasing — the link's `aria-label` (`"Restricted article (financial, internal)"`) *and* the icon's `role="img" aria-label` (`"Restricted: financial, internal"`). In practice the link's `aria-label` wins during link navigation, so the icon label is only encountered on intra-element traversal — harmless. If you ever want a single source, you could drop the icon's label and rely on `articleCardLabel` alone, but leaving it is defensible (belt-and-suspenders for the most security-relevant affordance on the page).

### Deliberately left (correct calls)
- **B5** (featured tag slice `3 → 4`) — intentional; not reverted. Fine.
- **M1** (`RestrictedArticleIcon` still inline / one call site) — trivial; fine for now.
- Broader pre-existing `--text-faint` usage (e.g. `.article-date`) — out of scope for this PR; a separate contrast pass if you want it.

Solid iteration. This is ready.

Issue #10, opened 2026-04-24
Implement Recall Orchestrator core


SweetSophia:
Part of #5\n\nBuild the central retrieval engine that queries enabled providers, normalizes results, and prepares them for downstream arbitration and budgeting.\n\n## Deliverables\n- multi-provider query fanout\n- result normalization\n- ranking pipeline hooks\n- output modes for prompt injection vs inspection

PR #1, opened 2026-04-11
Security hardening: fixes from code review

SweetSophia:
## Summary

Security hardening fixes based on comprehensive code review using security-auditor + requesting-code-review skills.

### Changes

1. **Health endpoint with DB connectivity check** (`src/app/api/health/route.ts`)
   - Now verifies database connectivity before returning 200
   - Returns 503 on failure

2. **Path traversal prevention** (`src/lib/uploads.ts`)
   - Enhanced `resolvePublicImagePath` with strict validation
   - Added directory boundary check to prevent path escape
   - Regex validation for allowed characters only

3. **Content length validation** (`src/app/api/articles/route.ts`)
   - Added `MAX_CONTENT_SIZE` (1MB)
   - Added `MAX_TITLE_LENGTH` (200 chars)
   - Added `MAX_EXCERPT_LENGTH` (500 chars)
   - Validates content size before storing

4. **Reusable auth middleware** (`src/lib/api/auth.ts`)
   - Created `checkRouteAuth()`, `hasPermission()`, `requirePermission()` utilities
   - Reduces duplicated permission checking logic across routes

### Issues Addressed

| Severity | Issue | Status |
|----------|-------|--------|
| Critical | No content-length validation | ✅ Fixed |
| Important | Path traversal in image resolution | ✅ Fixed |
| Minor | Health endpoint returns 200 without DB check | ✅ Fixed |

### Verification

- ✅ No eval() usage
- ✅ No innerHTML/dangerouslySetInnerHTML
- ✅ API keys properly hashed (SHA-256)
- ✅ Prisma parameterized queries (no SQL injection)
- ✅ Environment variables used safely

Full review report: `CODE_REVIEW_REPORT.md`

## Summary by CodeRabbit

* **New Features**
  * Centralized API authorization with role/permission checks
  * Search supports filtering by status and confidence and preserves result ordering when using external search

* **Bug Fixes**
  * Health endpoint detects DB connectivity issues and returns 503 on failure
  * Search listing and search results are consistently filtered

* **Security**
  * Enforced article input size and length limits with early validation
  * Stricter upload path validation to prevent traversal

* **Other**
  * Tags: array/type validation, normalization, deduplication, upserted and standardized in responses

Issue #14, opened 2026-04-24
Add user-configurable recall settings


SweetSophia:
Part of #5\n\nExpose recall behavior settings to the user.\n\n## Settings\n- autoRecallEnabled\n- maxInjectedMemories\n- maxInjectedTokens\n- recallVerbosity\n- enabledProviders\n- providerPriorityWeights

PR #222, opened 2026-06-26
refine wiki topic page

SweetSophia:
## Refined Description

This PR refines the wiki topic page as part of the PR-3 revamp phase, focusing on hierarchy clarity, accessibility, and information density for both desktop and mobile users.

### Key Changes

**Breadcrumb & Navigation**
- Replaced the single-parent breadcrumb with a full ancestor path traversal (`buildTopicPath`) so deeply nested topics display their complete hierarchy from Noosphere root to the current topic
- The current topic is now rendered as a non-link terminal breadcrumb, with ancestors linking to their respective pages

**Hero & Topic Metadata**
- Added a `topic-page-hero` class for shared hero styling with the wiki home page
- Introduced a "Root topic branch" pill that displays when a topic has no parent but has subtopics, providing context for top-level branches
- Added role-aware create actions (article + subtopic) gated on user permissions

**Subtopic Cards**
- Subtopics are now ordered alphabetically (`orderBy: { name: "asc" }`)
- Each subtopic card now shows **two metrics**: direct article count and child subtopic count, displayed side-by-side in a bordered grid for easier scanning
- Aria-labels describe the full metric composition for screen readers (e.g., `"Serianis Project: 5 direct articles, 2 child subtopics"`)

**Article Cards**
- Extracted `RestrictedArticleIcon` as a reusable component with proper `role="img"` and `aria-label` (replacing the previous `title` tooltip)
- Added a dedicated `articleCardLabel` builder to produce comprehensive aria-labels including restriction status, confidence, and author
- Moved status (`status-badge`) and confidence (`confidence-badge`) into a right-aligned signals column, creating a clearer visual separation between content and metadata
- Capped visible tags at 5 with an explicit italic `+N more` overflow chip for articles with many tags
- Dates now format consistently using an `Intl.DateTimeFormat` instance (`en-GB`, e.g., "12 Mar 2024")

**CSS / Responsive Layout**
- Extended shared CSS patterns: `.topic-page-hero`, `.topic-page-section`, `.topic-subtopic-metrics`, `.topic-article-card`, `.topic-article-main`, `.topic-article-signals`, and `.tag-badge-more`
- Article cards collapse to single-column on mobile with signals left-aligned (no horizontal overflow)
- Subtopic metrics grid stacks gracefully on narrow viewports
- Text in article card body elements uses `overflow-wrap: anywhere` to prevent tag/badge overflow

---

### Summary

Refined the wiki topic page and related home page to extract shared helpers into a dedicated formatting module, improve accessibility, and tighten the visual layout.

### Changes

- **Extracted shared helpers into `src/lib/wiki-format.ts`**
  - Centralized the `wikiDateFormatter` (now UTC-based), `pluralize`, and `articleCardLabel` helpers, removing duplicate inline definitions previously scattered across the topic and home pages.
  - `articleCardLabel` now produces a more descriptive, screen-reader-friendly label that includes the topic name, author, status, and confidence when available, and uses a structured `Status:` / `Confidence:` suffix for clearer assistive tech output.

- **Extracted `RestrictedArticleIcon` into a reusable component (`src/components/wiki/RestrictedArticleIcon.tsx`)**
  - Pulled the lock icon out of both the topic and home page modules so it can be reused.
  - Replaced the previous `role="img"` + `aria-label` pattern with `aria-hidden="true"` plus a native `title` tooltip, avoiding redundant screen reader announcements.

- **Updated `src/app/wiki/[topicSlug]/page.tsx`**
  - Imports the new shared helpers and component.
  - Optimized the Prisma `topic.findUnique` query to select only the parent's `name` instead of fetching the full record.
  - Removed unnecessary ordering on the global topic fetch (the page already sorts locally).
  - Refined copy: "Root / topic branch" → "Root / top-level topic" and rewrote the Subtopics section subtitle to better describe the next-level branches.
  - Improved subtopic card semantics and screen reader labels (prefixed with "Subtopic", prefixed counts with "direct", used shared `pluralize`).
  - Swapped the in-card CSS class from `topic-tree-kind` to a new `subtopic-kind` class with dedicated styling.

- **Updated `src/app/wiki/page.tsx`**
  - Imports the shared helpers and icon component, removing duplicate local definitions.
  - Marks the muted article tag row as `aria-hidden="true"` so its tags aren't redundantly announced alongside the card label.

- **CSS refinements in `src/app/wiki/wiki.css`**
  - `.topic-subtopic-card` now uses an explicit `minmax(0, 1fr) auto` grid for predictable two-column behavior, with a dedicated `.topic-subtopic-copy` container to allow text to wrap properly.
  - Added `.subtopic-kind` styling for the small uppercase "Subtopic" kicker.
  - Applied `overflow-wrap: break-word` to `.topic-article-main h3` and `.topic-article-main p` to prevent long titles or excerpts from overflowing.
  - Tweaked the responsive rules so `.sub-topic-tag` and `.tag-badge` keep `justify-content: center` while regaining flexible width, and ensured `.page-meta-pill` centers its content on small screens.

### Impact

These changes primarily improve maintainability (shared formatting helpers, reusable icon component), accessibility (cleaner aria-label text, redundant icon labels removed, decorative content hidden), and visual consistency on the wiki topic and home pages.

### DIFF
--- a/src/app/wiki/[topicSlug]/page.tsx
+++ b/src/app/wiki/[topicSlug]/page.tsx
@@ -13,6 +13,79 @@ interface Props {
   params: Promise<{ topicSlug: string }>;
 }
 
+interface TopicPathNode {
+  id: string;
+  name: string;
+  slug: string;
+  parentId: string | null;
+}
+
+const topicDateFormatter = new Intl.DateTimeFormat("en-GB", {
+  month: "short",
+  day: "numeric",
+  year: "numeric",
+});
+
+function pluralize(count: number, singular: string, plural = `${singular}s`) {
+  return count === 1 ? singular : plural;
+}
+
+function buildTopicPath(topics: TopicPathNode[], current: TopicPathNode) {
+  const topicMap = new Map(topics.map((topic) => [topic.id, topic]));
+  const path: TopicPathNode[] = [];
+  const seen = new Set<string>();
+  let cursor: TopicPathNode | undefined = current;
+
+  while (cursor && !seen.has(cursor.id)) {
+    path.unshift(cursor);
+    seen.add(cursor.id);
+    cursor = cursor.parentId ? topicMap.get(cursor.parentId) : undefined;
+  }
+
+  return path;
+}
+
+function RestrictedArticleIcon({ tags }: { tags: string[] }) {
+  if (tags.length === 0) return null;
+
+  return (
+    <span className="restricted-icon" role="img" aria-label={`Restricted: ${tags.join(", ")}`}>
+      <svg
+        width="12"
+        height="12"
+        viewBox="0 0 24 24"
+        fill="none"
+        stroke="currentColor"
+        strokeWidth="2"
+        strokeLinecap="round"
+        strokeLinejoin="round"
+        aria-hidden="true"
+      >
+        <rect x="3" y="11" width="18" height="11" rx="2" ry="2" />
+        <path d="M7 11V7a5 5 0 0 1 10 0v4" />
+      </svg>
+    </span>
+  );
+}
+
+function articleCardLabel(article: {
+  title: string;
+  author?: { name: string | null } | null;
+  authorName: string | null;
+  status: string;
+  confidence: string | null;
+  restrictedTags: string[];
+}) {
+  const author = article.author?.name ?? article.authorName ?? "Unknown author";
+  const restriction =
+    article.restrictedTags.length > 0
+      ? `Restricted article (${article.restrictedTags.join(", ")})`
+      : "Article";
+  const confidence = article.confidence ? `, ${article.confidence} confidence` : "";
+
+  return `${restriction}: ${article.title}, ${article.status}${confidence}, by ${author}`;
+}
+
 export default async function TopicPage({ params }: Props) {
   const { topicSlug } = await params;
 
@@ -26,8 +99,9 @@ export default async function TopicPage({ params }: Props) {
       parent: true,
       children: {
         include: {
-          _count: { select: { articles: true } },
+          _count: { select: { articles: true, children: true } },
         },
+        orderBy: { name: "asc" },
       },
     },
   });
@@ -35,6 +109,12 @@ export default async function TopicPage({ params }: Props) {
   if (!topic) {
     notFound();
   }
+
+  const allTopics = await prisma.topic.findMany({
+    select: { id: true, name: true, slug: true, parentId: true },
+    orderBy: { name: "asc" },
+  });
+  const topicPath = buildTopicPath(allTopics, topic);
 
   // Unauthenticated users only see unrestricted articles.
   // Human sessions (via NextAuth) always have full access — they bypass restrictions.
@@ -49,14 +129,17 @@ export default async function TopicPage({ params }: Props) {
     },
     orderBy: { updatedAt: "desc" },
   });
+  const hasSubtopics = topic.children.length > 0;
 
   return (
     <div className="wiki-content topic-page">
       <Breadcrumbs
         items={[
           { label: "Noosphere", href: "/wiki" },
-          ...(topic.parent ? [{ label: topic.parent.name, href: `/wiki/${topic.parent.slug}` }] : []),
-          { label: topic.name },
+          ...topicPath.map((pathTopic, index) => ({
+            label: pathTopic.name,
+            href: index === topicPath.length - 1 ? undefined : `/wiki/${pathTopic.slug}`,
+          })),
         ]}
       />
 
@@ -64,6 +147,7 @@ export default async function TopicPage({ params }: Props) {
         eyebrow="Topic"
         title={topic.name}
         description={topic.description ?? "A focused collection of articles and subtopics inside the Noosphere knowledge graph."}
+        className="topic-page-hero"
         actions={
           canCreateArticle ? (
             <Link href={`/wiki/${topic.slug}/new`} className="btn btn-primary btn-sm">
@@ -86,32 +170,48 @@ export default async function TopicPage({ params }: Props) {
                 <strong>{topic.parent.name}</strong>
                 <span>parent topic</span>
               </span>
+            ) : hasSubtopics ? (
+              <span className="page-meta-pill">
+                <strong>Root</strong>
+                <span>topic branch</span>
+              </span>
             ) : null}
           </div>
         }
       />
 
-      {topic.children.length > 0 && (
-        <section className="browse-section browse-section-tight">
+      {hasSubtopics && (
+        <section className="browse-section browse-section-tight topic-page-section">
           <div className="section-header">
             <div className="section-header-copy">
               <p className="page-eyebrow">Branch outward</p>
               <h2 className="section-title">Subtopics</h2>
-              <p className="section-subtitle">Jump into narrower branches connected to this topic.</p>
+              <p className="section-subtitle">Move into narrower branches without losing the local article context.</p>
             </div>
           </div>
 
           <div className="topic-subtopic-grid">
             {topic.children.map((child) => (
-              <Link key={child.id} href={`/wiki/${child.slug}`} className="topic-card topic-subtopic-card">
+              <Link
+                key={child.id}
+                href={`/wiki/${child.slug}`}
+                className="topic-card topic-subtopic-card"
+                aria-label={`${child.name}: ${child._count.articles} direct ${pluralize(child._count.articles, "article")}, ${child._count.children} child ${pluralize(child._count.children, "subtopic")}`}
+              >
                 <div>
-                  <p className="topic-tree-kind">Subtopic</p>
+                  <p className="topic-tree-kind" aria-hidden="true">Subtopic</p>
                   <h3>{child.name}</h3>
                   <p>{child.description ?? "Explore the next layer of articles nested under this branch."}</p>
                 </div>
-                <div className="topic-subtopic-count">
-                  <strong>{child._count.articles}</strong>
-                  <span>article{child._count.articles !== 1 ? "s" : ""}</span>
+                <div className="topic-subtopic-metrics" aria-hidden="true">
+                  <div className="topic-subtopic-count">
+                    <strong>{child._count.articles}</strong>
+                    <span>direct article{child._count.articles !== 1 ? "s" : ""}</span>
+                  </div>
+                  <div className="topic-subtopic-count">
+                    <strong>{child._count.children}</strong>
+                    <span>child subtopic{child._count.children !== 1 ? "s" : ""}</span>
+                  </div>
                 </div>
               </Link>
             ))}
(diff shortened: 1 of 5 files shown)


SweetSophia:
## Review — PR #222 (`refine wiki topic page`)

**Verdict:** Ship after the three critical a11y fixes and the helper-extraction. No behavior regressions. Several real-but-non-blocking improvements worth filing as follow-ups.

I reviewed the diff, the full file (`src/app/wiki/[topicSlug]/page.tsx`), `wiki.css`, the sibling files (`Breadcrumbs`, `PageHeader`, `src/app/wiki/page.tsx`, `prisma/schema.prisma`), ran local `tsc --noEmit` (clean) and `eslint` (clean), and reconciled against the Kody / kilo / Gemini / Codex reviews already on the PR.

---

### 🔴 Critical — must-fix before merge

**1. `articleCardLabel` reads "low confidence, by Alice" out loud**
`src/app/wiki/[topicSlug]/page.tsx:84–86`

Current output:
> "Restricted article (financial): Q3 Report, draft, low confidence, by Cylena"

Two parse failures for a screen-reader user:
- "low confidence" sits between status and author → sounds like Cylena has low confidence.
- The `(financial)` parenthetical has nothing to scope it; "financial: My Article" mis-parses as a label.

Suggested phrasing (one helper, period-separated clauses):
```ts
const parts = [
  `${restriction}: ${article.title}, by ${author}.`,
  `Status: ${article.status}.`,
  article.confidence ? `Confidence: ${article.confidence}.` : null,
].filter(Boolean);
return parts.join(" ");
```
Result: `"Restricted article (financial): Q3 Report, by Cylena. Status: draft. Confidence: low."`

**2. `RestrictedArticleIcon` `role="img"` double-announces the restriction**
`src/app/wiki/[topicSlug]/page.tsx:52`

The icon's wrapper has `role="img"` + `aria-label="Restricted: financial"`, AND the parent `<Link>` has `aria-label="Restricted article (financial): …"`. Screen readers announce both — same info, different word order. The icon is decorative inside a labelled link; drop the `role="img"` and `aria-label` on the wrapper, keep `aria-hidden="true"` on the inner `<svg>`. Add a `title=` on the wrapper for sighted-hover discoverability.

**3. `overflow-wrap: anywhere` will fragment tag text mid-word**
`src/app/wiki/wiki.css:874–878`

`anywhere` may break ANY character — including inside a single tag like `#incident-response` rendering as `#incident-` / `response`. Use `break-word` for prose, keep `anywhere` only for `.tag-badge` where arbitrary user input lives:
```css
.topic-article-main h3,
.topic-article-main p { overflow-wrap: break-word; }
.topic-article-main .tag-badge { overflow-wrap: anywhere; }
```
(Also flagging because the `.tag-badge` rule is what kilo's review hit but the prose rule was the actual dangerous one.)

---

### 🟡 Important — should-fix

**4. `pluralize()` is defined and then ignored four times**
`src/app/wiki/[topicSlug]/page.tsx:29–31` defines it. `page.tsx:162, 166, 209, 213` all inline `!== 1 ? "s" : ""` instead. This is the worst-of-both-worlds — an abstraction that doesn't pay for itself. Either delete `pluralize` (YAGNI) or use it at every call site. I recommend keeping it: the maintenance burden of "did I remember to switch?" is real and the helper already supports a custom plural.

**5. `articleCardLabel`, `pluralize`, `topicDateFormatter`, and `RestrictedArticleIcon` are duplicated between this file and `src/app/wiki/page.tsx`**
- `RestrictedArticleIcon` — verbatim SVG copy (`[topicSlug]/page.tsx:48–69` ≈ `page.tsx:55–76`)
- `pluralize` — verbatim (`[topicSlug]/page.tsx:29–31` ≈ `page.tsx:86–88`)
- `topicDateFormatter` — same config (`[topicSlug]/page.tsx:23–27` ≈ `page.tsx:49–53`)
- `articleCardLabel` — same shape, divergent signatures (this PR's version adds status + confidence, home page's version adds topic name)

Extract to:
- `src/components/wiki/RestrictedArticleIcon.tsx`
- `src/lib/format.ts` (formatter + `pluralize`)
- One shared `articleCardLabel` taking a normalized record, with the home page passing `article.topic?.name` as an additional field.

Without this, any of the fixes #1, #2 above have to land in two files and stay in sync.

**6. `_count.children` returns total descendants, not direct children — label is misleading**
`src/app/wiki/[topicSlug]/page.tsx:102, 213`

Prisma's `_count.children` counts through the entire `children` relation — i.e., the whole subtree, not direct children. The visible label says "child subtopic" and the aria-label says "child subtopic(s)". Both understate. Either rename to "subtopic" (matches home page's `descendantCount` convention) or compute direct children from the already-fetched `topic.children` array.

**7. `Intl.DateTimeFormat` has no `timeZone` → SSR hydration risk**
`src/app/wiki/[topicSlug]/page.tsx:23–27`

No `timeZone` option means the formatter uses the runtime's local timezone. Server (likely UTC) formats `article.updatedAt` differently than a client in a non-UTC timezone, which can shift the displayed day. For a date stored as `2024-03-12T01:30:00Z`, server renders "12 Mar 2024" while an EST client re-renders "11 Mar 2024" → hydration mismatch warning in dev, silent inconsistency in prod. Add `timeZone: "UTC"` (or any explicit zone) to lock formatting.

---

### 🟢 Polish — nice-to-have

**8. Subtopic card `aria-label` lacks the "Subtopic" prefix and reads oddly in singular**
`src/app/wiki/[topicSlug]/page.tsx:199`

> `"Foo: 3 direct articles, 2 child subtopics"` vs `"Foo: 1 direct article, 1 child subtopic"`

"1 child subtopic" is grammatically broken (adjective without a noun). And unlike the article labels (which all start with "Article:" / "Restricted article:"), this one starts with the bare name. Suggest:
> `"Subtopic Foo: 3 direct articles, 2 nested subtopics"`

**9. "Root" / "topic branch" copy contradicts the tree metaphor**
`src/app/wiki/[topicSlug]/page.tsx:174–177`

A "branch" in a tree is a sub-tree — i.e., a node with children. A top-level node that IS the trunk isn't a branch. Pair like the other pills:
- `<strong>Root</strong><span>top-level topic</span>` (most explicit)
- or just `<strong>Root</strong><span>topic</span>` (parallels `parent topic`)

**10. `parent: true` include is half-redundant after `allTopics`**
`src/app/wiki/[topicSlug]/page.tsx:99, 168–172`

The parent pill uses only `topic.parent.name`; `topic.parent.slug` is no longer referenced anywhere. Now that `allTopics` is in scope, `buildTopicPath` already builds a `topicMap` — expose the immediate parent from there and drop the Prisma `parent: true` include (or narrow it to `select: { name: true }`).

**11. Subtopic cards stack vertically on desktop — feels like a feed entry, not a nav card**
`src/app/wiki/wiki.css:805–812`

`topic-subtopic-card` has no `grid-template-columns`, so the description always sits above the metrics row. On a 1280px viewport this gives a tall card. A `minmax(0, 1fr) auto` two-column layout (matching `.topic-article-card`) would balance better — description left, metrics right. Make sure to put `min-width: 0` on the description block.

**12. Section subtitle "Move into narrower branches without losing the local article context" is vague**
`src/app/wiki/[topicSlug]/page.tsx:189`

"local article context" is undefined and unclear why it would be lost. Suggest: `"Subtopics nested under this branch — open any to see its articles and nested branches."` Or simpler: `"Narrower branches connected to this topic."` (matches the original).

**13. `topic.parent` include over-fetches — only `name` is used**
`src/app/wiki/[topicSlug]/page.tsx:99`

`parent: true` selects every parent column (`description`, `createdAt`, `updatedAt`, `parentId`, …) when the meta pill needs only `name`. Either narrow with `select: { name: true }` or, with fix #10, drop the include entirely.

**14. Tags become full-width on ≤640px**
`src/app/wiki/wiki.css:2309–2314`

`.tag-badge { width: 100%; justify-content: center; }` on tiny screens stacks 5 tags + a `+N more` chip as six centered full-width rows. Drowns the italic `+N more` chip. Constrain to `.page-meta-pill` (the heavy pill), or threshold tighter at 480px with `align-self: flex-start` for tags.

**15. `.topic-tree-kind` reused for "Subtopic" kicker on a different layout**
`src/app/wiki/[topicSlug]/page.tsx:202`

The class is named for the wiki-home topic tree (`wiki.css:736–748`). Reusing it on subtopic cards creates cross-page styling coupling under a misleading name. Either accept the coupling or rename to `.subtopic-kind`.

**16. `orderBy: { name: "asc" }` on the topic `findMany` is wasted**
`src/app/wiki/[topicSlug]/page.tsx:113–116`

`buildTopicPath` only uses the rows as a `Map`; the order is never consumed. Drop the `orderBy` (it forces a sort without an index, per Oracle's I3 below — adding a `name` index would solve it, but it's still a pointless order-by for this consumer).

---

### 🧱 Architectural follow-ups (file as separate issues)

**A1. Cache the full-table `prisma.topic.findMany`**

`page.tsx:113–116` runs a full topic scan on every `force-dynamic` render. Fine at current seed (~15 topics), fine at 1k, painful at 10k. Two improvements:
1. Add `@@index([name])` to the Topic model — currently only `@@index([parentId])` and `@@index([slug])` exist (`schema.prisma:18–30`).
2. Wrap with `unstable_cache` tagged `["topics"]` + `revalidateTag("topics")` from the admin topic create/update/delete actions (which already call `revalidatePath("/wiki")` per `src/app/wiki/admin/topics/actions.ts`).

`React.cache()` (request-scoped) is NOT useful here — it's a single call per render.

**A2. `TopicPathNode` interface duplicates Prisma's generated type**

`page.tsx:16–21` hand-declares `{ id, name, slug, parentId }` — exactly what `Prisma.TopicGetPayload<{ select: { id: true; name: true; slug: true; parentId: true } }>` produces for free, and stays in sync with schema migrations.

**A3. Empty-state doesn't distinguish a leaf topic with zero content from a leaf with zero articles**

If a topic has no subtopics and no articles, the user sees a hero + a "No articles yet" block — no signal they're at the bottom of the tree, no quick way up beyond the breadcrumb. Either hide the Articles eyebrow entirely in that case, or surface a "topic has no content yet" state with a parent breadcrumb as a call-to-action.

---

### ✅ What's done well

- **`buildTopicPath` cycle protection** — `seen` set silently recovers from a malformed parent chain instead of 500ing. Defensive but cheap.
- **Decorative elements consistently `aria-hidden`** — kicker, date, tag row, subtopic metrics, signals column all correctly hidden because the parent `aria-label` carries the info. Clean pattern.
- **`+N more` tag truncation** — cap of 5 + overflow chip is more usable than truncating.
- **`hasSubtopics` local var** — small readability win over three `topic.children.length > 0` re-evaluations.
- **`result-count` only renders when articles exist** — avoids the silly "0 total" badge.
- **Alphabetical subtopic sort** — predictable scan order.
- **Mobile collapse is well-defined** — at 960px, signals drop `min-width: 8.5rem` and go left-aligned; subtopic metrics collapse to single column with `text-align: left`. Both graceful.
- **`buildScopeFilter` reuse** — consistent access control with the home page.
- **Module-level `topicDateFormatter` singleton** — correct pattern (per Oracle); just needs the `timeZone` lock.

---

### 🤝 Reconciliation with existing bot reviews

- **Kody** ("no issues") — missed all of the above; Kody's ruleset doesn't appear to cover a11y phrasing, CSS typography, or Prisma query shape.
- **kilo** — caught the full-table scan (A1) and `overflow-wrap` (#3 above). Missed the `parent: true` redundancy, the aria-label phrasing (#1), the icon double-announcement (#2), the `_count.children` semantics (#6), the SSR TZ bug (#7), and the duplicated helpers (#5).
- **Gemini** — flagged the perf scan (A1) and mobile subtopic metrics stacking (#14). Missed the rest.
- **Codex** — no specific suggestions.

PR #233, opened 2026-06-28
Add markdown editor preview modes

SweetSophia:
## Summary

Adds a consolidated wiki markdown editor with Write, Preview, and Split modes inside `MarkdownPreviewTabs`.

## Current Behavior

- `MarkdownPreviewTabs` owns the toolbar, controlled textarea, hint, preview pane, and ARIA radiogroup.
- The textarea remains the submitted `name=\"content\"` field for the existing create/edit server actions.
- Required empty content cannot switch into Preview/Split; the component focuses the textarea and calls `reportValidity()` while keeping the editor in Write mode.
- Edit pages reset editor state through `key={article.id}` remounting. New article pages do not use a topic key because their content starts empty.
- Toolbar and image-upload insertions sync through `MARKDOWN_TEXTAREA_UPDATE_EVENT`; image insertion now mirrors toolbar block spacing around existing text.
- Split mode falls back from current React content state when the viewport no longer supports split layout.

## Accessibility

- Mode controls use `role=\"radiogroup\"` / `role=\"radio\"`, roving `tabIndex`, `aria-checked`, `aria-orientation=\"horizontal\"`, and keyboard support for arrows, Home, End, and Space.
- All mode buttons use one stable `aria-controls` target: the editor shell.
- Split mode uses `aria-describedby` with visually hidden text instead of `aria-description`.

## Testing

- Added DOM render coverage for `MarkdownPreviewTabs` with `@testing-library/react` and `happy-dom`.
- Covered ARIA output, roving keyboard behavior, required-empty preview blocking, controlled toolbar sync, selection restoration, and split fallback behavior.

# Add markdown editor preview modes

This pull request introduces and refines a tabbed markdown editor component that lets users switch between Write, Preview, and Split (side-by-side) modes for editing wiki content.

## Highlights

- **New `MarkdownPreviewTabs` component** with an ARIA radiogroup for Write / Split / Preview modes, a roving-tabindex keyboard toolbar, and a stable `*-shell` element so screen readers can target the active mode.
- **Responsive fallback to stacked modes** when the viewport is too narrow for a side-by-side layout, falling back to Preview (when there is content) or Write when Split mode is no longer available.
- **Required-field safety net**: switching to Preview with an empty required editor re-focuses the textarea and surfaces its native validation state, preventing users from getting stuck.
- **Spacebar support** on the radiogroup activates the focused mode without scrolling the page.
- **Image upload integration**: `ImageUploadPanel` now uses `updateMarkdownTextarea` to insert markdown images with proper paragraph spacing, falling back to a single newline when adjacent content already ends/starts on a newline.
- **Controlled toolbar updates**: `updateMarkdownTextarea` now consistently restores the requested selection after toolbar-driven mutations (tested via the bold-wrapper flow), making the new tabs + image upload combo behave predictably.
- **Accessibility polish**: `.visually-hidden` utility now also uses `clip-path: inset(50%)` (in addition to the legacy `clip` rule) for modern browsers.
- **Server-component fix**: removed the per-request `key={topicSlug}` from the new article page so the editor mounts cleanly without remounting between renders.

## Tests

- New test suite (`src/__tests__/wiki/markdown-preview-tabs.test.ts`, 5 tests) covers:
  - ARIA radiogroup structure and shared `aria-controls` pointing at the editor shell.
  - Roving keyboard navigation (ArrowRight, End, Home) and that Space is `preventDefault`-ed.
  - Required-content validation when leaving Write mode empty.
  - Toolbar-driven text wrapping via `updateMarkdownTextarea` preserving the cursor.
  - Automatic fallback from Split to Preview when the media query no longer matches.

---

This PR adds markdown editor preview modes functionality. The change ensures that when switching between different topics in the wiki article creation flow, the markdown preview tabs properly reset their state by using the `topicSlug` as a React `key` prop. This forces a fresh component instance per topic, preventing stale preview state from carrying over when users navigate to create articles under different topics.


SweetSophia:
## Review — PR #233 (`polish/markdown-editor-preview-modes`)

**Verdict:** Ship after 4 fixes (1 real bug, 1 invariant, 1 a11y, 1 UX). Architecture is sound; the radiogroup, hoisted plugins, and `key`-based defaultValue sync are all clean wins. I reviewed the diff, both page files, `MarkdownPreviewTabs.tsx`, `MarkdownToolbar.tsx`, the new CSS, ran local `tsc --noEmit` (clean) and `eslint` (clean), and reconciled against Kody / kilo / Gemini / Codex reviews + an independent oracle + designer pass.

---

### 🔴 Critical — must-fix before merge

**1. New page is missing `key={topicSlug}` — cross-topic content bleed**
`src/app/wiki/[topicSlug]/new/page.tsx:100–106`

The edit page (`edit/page.tsx:137`) has `key={article.id}` for exactly this reason. The new page has the same bug, just keyed against `topicSlug` instead.

**Trace:** User opens `/wiki/engineering/new`, types "Here's how the build pipeline works.", clicks a Link to `/wiki/security/new`. App Router re-renders the server component with new params. The client component `MarkdownPreviewTabs` is at the same tree position with no key, so React reuses the instance. `useState(defaultValue ?? "")` was initialized on first mount with `""` — it does NOT re-run. `content` retains the engineering text. User then creates an article in **security** with engineering's body.

This is **worse** than the edit page case the PR already fixes: it surfaces after a single nav between topics, a more common workflow than editing two articles back-to-back.

Fix:
```tsx
<MarkdownPreviewTabs
  key={topicSlug}
  targetTextareaId="content"
  ...
/>
```

**2. `required={required && activeMode !== "preview"}` is a load-bearing invariant**
`MarkdownPreviewTabs.tsx:154`

The gating logic (`canHideEditor`, line 71) correctly blocks entering preview mode with empty required content. So at the moment you click "Preview", the invariant `required && empty → can't be here` holds. But the invariant is **undocumented and fragile** — it depends on the user not being able to edit content in preview mode, which is enforced only by CSS `display: none` at `wiki.css:1279–1281`.

If anyone ever adds an "Edit in preview" affordance, content edit shortcut, or any way for the textarea to become user-editable while `data-mode="preview"`, the form silently accepts empty content. And the conditional `required` means HTML form validation in preview mode silently passes an empty body — server-side validation in `saveArticle` / `createArticle` becomes load-bearing in a way that's easy to miss.

Fix: keep `required={required}` always. The browser correctly validates `required` fields even when `display: none` — that's standard HTML. Then `canHideEditor` becomes purely a UX gate, not a correctness gate. (Please verify `src/app/wiki/[topicSlug]/new/actions.ts` and `[articleSlug]/edit/actions.ts` do server-side content validation as a defense-in-depth.)

**3. Silent failure on keyboard-blocked preview switch — a11y regression**
`MarkdownPreviewTabs.tsx:73–89`

When a keyboard user presses → to move from "Write" to "Preview" while the textarea is empty and required, the radiogroup silently does nothing. The `event.preventDefault()` already consumed the arrow press, focus stays on "Write", the preview never opens, the user has no signal. Click gives native `reportValidity()` feedback (focus + browser tooltip). Keyboard gives nothing.

This violates WCAG 3.3.1 (Error Identification) and breaks the radiogroup contract — a user moving through radios with arrows expects *something* to happen, even if blocked.

Fix (symmetric with click path):
```tsx
if (mode === "preview" && !canHideEditor) {
  const textarea = textareaRef.current;
  textarea?.focus();
  textarea?.reportValidity();
  return;
}
```
Or, if you want to keep focus on the radiogroup, add a `aria-live="polite"` span with an announcement.

**4. Split mode on mobile = ~56rem of stacked panes**
`wiki.css:1264–1301`

Below 980px, the grid collapses to single column but **both** `.markdown-editor-textarea` and `.markdown-editor-preview` keep `min-height: 28rem`. So on a phone the editor is ~950–1000px tall of duplicated source/target views that scroll independently. A user has to scroll an enormous distance to see both views of the same content.

Recommendation: make Split desktop-only. Below 980px, hide the Split button via CSS and force fallback to Preview (or auto-downgrade in `selectMode` when `window.matchMedia("(min-width: 980px)")` doesn't match).

```css
@media (max-width: 979px) {
  .preview-tab[data-mode="split"] { display: none; }
}
```

---

### 🟡 Important — should-fix

**5. Manual `input` event listener is functionally redundant**
`MarkdownPreviewTabs.tsx:59–68`

The textarea is controlled (`value={content}`) AND has both:
- `onChange={(event) => setContent(event.currentTarget.value)}` (line 156)
- A native `input` listener that calls `setContent(textarea.value)` (lines 66–68)

In React 17+ (this project uses React 19), the toolbar's `dispatchEvent(new Event("input"))` bubbles through React's delegated listener and fires `onChange` normally. The manual listener does exactly what React's onChange already does — two identical `setContent` calls per keystroke/toolbar-click, deduped silently.

The comment on line 65 ("Toolbar and image insertion update the DOM textarea directly, then emit input") implies the listener exists because React's onChange wouldn't fire. **That's not accurate in React 17+** — it's a vestigial safety net. It also masks a subtle bug surface: if React's onChange ever stops firing (e.g., someone replaces the dispatch with `setNativeValue`), the manual listener becomes load-bearing without anyone knowing.

Fix: delete lines 59–68. Keep only React's onChange. Add an integration test that exercises the toolbar flow (Bold/Link/Code/Image) to catch any regression.

**6. Empty-preview copy is the same in Preview and Split**
`MarkdownPreviewTabs.tsx:169`

In Preview mode "Nothing to preview yet." is appropriate — the user explicitly asked to see rendered output and there isn't any. In Split mode it's redundant: the textarea is right next to it with a placeholder. Two empty states side-by-side is clutter.

Fix:
```tsx
{activeMode === "write" ? null : trimmedContent ? (
  <ReactMarkdown ...>{trimmedContent}</ReactMarkdown>
) : activeMode === "split" ? (
  <p className="text-muted">Preview will appear here as you type.</p>
) : (
  <p className="text-muted">Nothing to preview yet.</p>
)}
```

**7. `aria-describedby` points into the (possibly hidden) write pane**
`MarkdownPreviewTabs.tsx:155, 159`

In Preview mode the write pane is `display: none`, which removes the hint `<p>` from the accessibility tree. The textarea's `aria-describedby` then resolves to nothing. In practice this is masked because the textarea is also `display: none` in Preview mode (not in tab order, not announced). But it's a latent issue if AT ever focuses the textarea directly.

Fix: gate the attribute on visibility:
```tsx
aria-describedby={hint && activeMode !== "preview" ? `${targetTextareaId}-hint` : undefined}
```

---

### 🟢 Polish — nice-to-have

**8. `aria-controls` with space-separated IDs for Split mode is partially honored**
`MarkdownPreviewTabs.tsx:111–114`

Spec-correct (IDREFS may be space-separated), but VoiceOver and TalkBack historically honor only the first ID. A Split-mode user on VoiceOver hears "controls content-write" with no signal that the preview is also under this radio's control.

Fix (pick one): drop `aria-controls` for Split and keep it only for Write/Preview (the simpler 1:1 relationships), OR add `aria-description="Shows the editor and rendered preview side by side"` on the Split button.

**9. Callback ref identity churn**
`MarkdownPreviewTabs.tsx:128–130`

```tsx
ref={(button) => { modeButtonRefs.current[mode] = button; }}
```
The arrow function is recreated each render. React calls the previous ref with `null` and the new ref with the same node. Harmless because the rAF-based reads (`selectMode` lines 76, 87) happen after reattachment, but wasteful.

Fix: useCallback the factory per mode. Minor optimization.

**10. `.preview-tab` CSS class is a grep-trap**
`wiki.css:1243`

The component is now a radiogroup, not tabs. Visual treatment is pill-shaped with `.active` — fine as styling. But the class name `preview-tab` and the file name `MarkdownPreviewTabs.tsx` mislead anyone reading the source.

Fix: rename `.preview-tab` → `.editor-mode-button`. Optionally rename the file to `MarkdownEditor.tsx` in a follow-up.

**11. Refs don't clear on unmount**
`MarkdownPreviewTabs.tsx:128–130`

Callback ref doesn't null out `modeButtonRefs.current[mode]` on unmount. Defensive hygiene. In practice the component is short-lived and the refs go out of scope with it. But if the component is ever rendered conditionally or in a list, stale refs could point to detached DOM nodes.

Fix: return a cleanup callback from the ref factory.

---

### 📌 Architectural follow-ups (file as separate issues — pre-existing, tangential)

**A1. `MarkdownToolbar` lacks roving tabindex**
`MarkdownToolbar.tsx:75–95`

Tab order is now: `[active mode radio] → [10 toolbar buttons] → [textarea]`. The radiogroup above uses correct roving tabindex; the toolbar immediately below does not. The inconsistency is jarring for keyboard users. Pre-existing, but the new well-designed radiogroup makes it more salient. Add roving tabindex in a follow-up.

**A2. `scroll-behavior: smooth` is not gated by `prefers-reduced-motion`**
`globals.css:311–314`

When the focus-management rAF in `selectMode` lands on a newly-active radio below the current scroll position, the browser smooth-scrolls. For users with `prefers-reduced-motion: reduce`, this animation can trigger nausea.

Fix:
```css
@media (prefers-reduced-motion: reduce) {
  html { scroll-behavior: auto; }
}
```

---

### ✅ What's done well

- **`REMARK_PLUGINS` / `REHYPE_PLUGINS` hoisted to module scope** — avoids re-creating arrays each render; preserves react-markdown's internal memoization.
- **`as const satisfies readonly EditorMode[]`** — combines literal-tuple autocomplete with compile-time exhaustiveness. Adding a mode to the union without updating the array is a type error. Excellent pattern.
- **Roving tabindex is correct** — only the active radio has `tabIndex={0}`; others get `-1`. APG-compliant.
- **`requestAnimationFrame` for focus moves** — correctly waits for React's render commit before focusing the newly-selected radio. Robust against React 18+ async rendering.
- **`event.preventDefault()` on arrow keys** — prevents page scroll from disrupting radiogroup navigation.
- **`Home` / `End` boundary keys** — APG-recommended.
- **`canHideEditor` gate before preview transition** — prevents the user from entering preview mode with empty required content.
- **`focusButton` option on `selectMode`** — differentiates keyboard (focus next button) from click (focus textarea + `reportValidity()`). Clean.
- **`required={required && activeMode !== "preview"}`** — keeps HTML validation from firing while the field is hidden. (But see finding #2: should be a UX-only relaxation, not a correctness one.)
- **Visual `data-mode` attribute** drives CSS visibility cleanly. Decouples "which panels are visible" from "which panel is controlled".
- **`key={article.id}` on the edit page** — the right answer for defaultValue reset. Simple, React-idiomatic, replaces the queueMicrotask + setContent dance from commits 1–2.
- **`name="content"` form contract preserved** across all modes (confirmed: `display:none` excludes from rendering only, not from form data submission).
- **The `useEffect(..., [])` deps on the input listener** — correct. Refs are stable, setContent is stable, sync closure captures the stable ref object.
- **Three commits iterated well** — the queueMicrotask approach was abandoned in favor of the cleaner `key` prop. Good engineering judgment to simplify.

---

### 🤝 Reconciliation with existing bot reviews

- **Kody** (3× "no issues") — missed all 4 critical findings. Kody's ruleset clearly doesn't cover a11y phrasing, radiogroup contracts, Next.js App Router client-component remount semantics, or CSS-only form-participation assumptions.
- **kilo** ("No Issues Found") — explicitly acknowledged the diff is clean but flagged nothing. Missed the new-page key gap, the required invariant, the redundant listener, and the silent keyboard failure.
- **Gemini** — acknowledged the radiogroup + split-mode refactor in a short summary, no specific issues raised.
- **Codex** — no specific suggestions.

SweetSophia:
## Deep review — PR #233

**Verdict:** Solid refactor with strong controlled-textarea discipline, a well-scoped `MARKDOWN_TEXTAREA_UPDATE_EVENT` protocol, and a correct ARIA radiogroup. A handful of targeted gaps before merge; no blockers. Lean toward merging after the test-coverage and image-insert gaps are closed.

### What works well
- **Controlled-textarea discipline.** Moved from a DOM `input`-event listener that mirrored into React state (the previous pattern) to a proper `value={content}` controlled textarea with a one-way `MARKDOWN_TEXTAREA_UPDATE_EVENT` sync from toolbar/upload → editor. This eliminates the stale-DOM and double-render races of the old pattern — a real architectural improvement, not just a refactor.
- **ARIA radiogroup.** Roving `tabIndex`, Arrow/Home/End nav with `requestAnimationFrame` focus restore, `aria-checked`, `aria-orientation="horizontal"`, `aria-describedby` + visually-hidden span for Split, and the `hint` is correctly suppressed in preview mode so the hidden parent isn't announced.
- **Defense against required+empty.** The `selectMode` gate (focus + `reportValidity()`) keeps the field on Write when the textarea is empty+required — the right place to enforce, instead of papering over with conditional `required`.

### Strong suggestions (worth fixing before merge)
1. **Add render tests for `MarkdownPreviewTabs` itself.** The new `markdown-textarea-update.test.ts` only covers the helper. The component has several untested load-bearing paths: roving tabindex on Arrow/Home/End, the required+empty preview gate, matchMedia resize fallback, ARIA attributes on the rendered output, and the `requestAnimationFrame` selection-restoration timing. Without these, future refactors in this high-traffic authoring surface could regress silently. Suggest `src/__tests__/wiki/markdown-preview-tabs.test.tsx` using `@testing-library/react` + `happy-dom`.
2. **Image-upload insert adds a leading blank line on empty articles** (`ImageUploadPanel.tsx:61`). `\`\n${result.markdown}\n\`` is used unconditionally, so when the cursor is at position 0 (very common on a fresh New Article) the inserted snippet produces a blank line at the top. The toolbar's `applyBlock` helper already handles this with a `spacerBefore` guard — mirror it for consistency.
3. **Resize fallback reads DOM, not state, for the "what mode to fall back to" decision** (`MarkdownPreviewTabs.tsx:90`). The rest of the file treats `content` as the source of truth; `textareaRef.current?.value.trim()` introduces a second one. Cache the latest trimmed content in a ref so Strict Mode double-invoke and just-queued `setContent` calls can't race.
4. **`aria-controls` on the Split radio is a two-ID IDREF list** (`MarkdownPreviewTabs.tsx:163`). WAI-ARIA permits it, but radiogroup controls are conventionally a single target and NVDA/JAWS often announce only the first id in a list. Either point all three radios at a new `.markdown-editor-shell` root id, or drop `aria-controls` from Write/Preview and keep it only on Split pointing at the grid container.

### Nits (optional, pick at your leisure)
- `.screen-reader-only` (`wiki.css:1239`) uses the legacy `clip: rect(0,0,0,0)` — add `clip-path: inset(50%)` for forward compat.
- The ref callback `(button) => { modeButtonRefs.current[mode] = button; }` recreates every render (`MarkdownPreviewTabs.tsx:157`); three named refs would be clearer.
- Space key isn't intercepted on the radios — `selectModeFromKeyboard` should include `" "` so Space doesn't trigger page scroll on browsers that scroll-button-on-Space.
- `key={topicSlug}` on the New Article page (`new/page.tsx:101`) is YAGNI — content is always empty there. Drop it. Keep `key={article.id}` on the Edit page (load-bearing for content reset).
- `MARKDOWN_TEXTAREA_UPDATE_EVENT` dispatches with `bubbles: true` but the only listener is on the textarea itself (`markdownTextareaUpdate.ts:34`). Either drop `bubbles` or add a one-line comment that it's intentional for future devtools/telemetry.

### Doc / code consistency
The PR description includes two claims that don't match the current code:
- "The textarea's `required` attribute is dynamically dropped in `preview` mode so HTML validation doesn't fire while the field is visually hidden." — actual code is unconditional `required={required}` (`MarkdownPreviewTabs.tsx:188`). Both are defensible (the gate at line 107 is the real enforcement, so the hidden field never actually blocks submit), but the description should be updated to match what's in the file.
- "defaultValue is re-applied via an effect when the prop changes" — the effect was removed in commit `6599643` ("Simplify markdown editor default content sync") in favor of the `key={article.id}` remount strategy. Description is stale.

### Inline comments
Pinning the three strongest actionable items (render-test gap, image-insert leading newline, resize fallback DOM read) below.

src/components/wiki/ImageUploadPanel.tsx
```diff
@@ -57,17 +58,11 @@ export function ImageUploadPanel({ targetTextareaId }: ImageUploadPanelProps) {
   function insertMarkdown() {
     if (!result) return;
 
-    const textarea = document.getElementById(targetTextareaId) as HTMLTextAreaElement | null;
-    if (!textarea) return;
-
     const snippet = `\n${result.markdown}\n`;
```
SweetSophia: Leading-blank-line bug on empty articles. `\n${result.markdown}\n` is wrapped unconditionally — when the cursor is at position 0 (very common on a fresh New Article) this produces a blank line at the top of the document. The toolbar's `applyBlock` already handles this with a `spacerBefore` guard:

```ts
const spacerBefore = start > 0 && value[start - 1] !== "\n" ? "\n\n" : "";
const inserted = `${spacerBefore}${result.markdown}\n`;
```

Mirror that here so image-upload behaves identically to toolbar block inserts. Could even factor a small `wrapAsBlock` helper if you want to share.

src/components/wiki/MarkdownPreviewTabs.tsx
```diff
@@ -15,70 +21,191 @@ const sanitizeSchema = {
...
-      textarea.removeEventListener("input", sync);
+  useEffect(() => {
+    const mediaQuery = window.matchMedia("(min-width: 980px)");
+    const syncSplitSupport = () => {
+      const matches = mediaQuery.matches;
+      setSupportsSplitMode(matches);
+      if (!matches) {
+        setActiveMode((currentMode) => {
+          if (currentMode !== "split") return currentMode;
+          return textareaRef.current?.value.trim() ? "preview" : "write";
```
SweetSophia: Two sources of truth. The rest of the file treats `content` state as the source of truth (line 100's `trimmedContent`); this fallback path reads `textareaRef.current?.value.trim()`. In normal flow they agree, but Strict Mode double-invoke and a just-queued `setContent` can race — the resize handler could observe stale DOM before the React commit lands. Suggest caching the latest trimmed content in a ref:

```ts
const trimmedContentRef = useRef("");
trimmedContentRef.current = trimmedContent;
// ...then in syncSplitSupport:
return trimmedContentRef.current ? "preview" : "write";
```

Drops the DOM read entirely.

src/components/wiki/MarkdownPreviewTabs.tsx
```diff
@@ -15,70 +21,191 @@ const sanitizeSchema = {
...
+        {EDITOR_MODES.map((mode) => (
+          <button
+            key={mode}
+            ref={(button) => {
+              modeButtonRefs.current[mode] = button;
+            }}
+            type="button"
+            role="radio"
+            aria-checked={activeMode === mode}
+            aria-controls={getModeControls(mode)}
```
SweetSophia: Two-ID IDREF list on `aria-controls`. `getModeControls('split')` returns `${writePaneId} ${previewPaneId}`. WAI-ARIA permits IDREF lists, but radiogroup controls are conventionally a single conceptual target, and NVDA/JAWS often announce only the first id in a list — Write and Preview correctly point at one pane, so the asymmetry also reads inconsistently to AT.

Cleanest fix: point all three radios at a single new shell id. Add `id={${targetTextareaId}-shell}` to the `.markdown-editor-shell` root and use `aria-controls={${targetTextareaId}-shell}` for every radio. Honest semantic relationship + no multi-ID edge case.

SweetSophia (commented):
Pinning three of the strongest actionable items from the consolidated review above (render-test gap, image-insert leading newline, resize fallback DOM read, aria-controls two-ID list). Each stands alone if you want to address them in separate commits.

SweetSophia:
## Follow-up review — PR #233 (commit `e1aad1a`)

**Verdict:** All 9 findings from the prior review correctly addressed, no regressions. **Ready to merge.** 6 minor nits below — feel free to address any of them or ship as-is; none are blocking.

### Fixes verified

| Prior finding | Status | Notes |
| --- | --- | --- |
| `aria-controls` pointed at one or two pane IDs (semantically inconsistent) | ✅ fixed | All three radios now point at a single `${targetTextareaId}-shell` container. Clean, defensible "this radio toggles a property on the shell" relationship. |
| Resize fallback read DOM (`textareaRef.current?.value.trim()`) — two sources of truth | ✅ fixed | Replaced with `trimmedContentRef`, populated by a `setContent` wrapper that mutates the ref synchronously before `setContentState`. Safe under StrictMode/concurrent rendering. |
| No render tests for `MarkdownPreviewTabs` | ✅ fixed | 5 new tests cover ARIA radiogroup, roving keyboard nav (incl. Space→preventDefault), required+empty Preview gate, controlled-textarea protocol + rAF selection restore, and matchMedia resize fallback. All 7 wiki tests pass locally (`npm run test:wiki`). |
| Selection-restoration rAF path untested | ✅ fixed | Test 4 asserts the post-condition (`selectionStart === 14, selectionEnd === 14`); passes because the contract holds both sync and async. |
| `ImageUploadPanel` leading newline bug | ✅ fixed | `insertMarkdown` now mirrors `applyBlock` exactly (line-by-line verified against `MarkdownToolbar.tsx:54-64`): `spacerBefore` skips when start===0 or prev char is `\n`; `spacerAfter` inserts `\n\n` only when needed. Behavior now matches the toolbar's Code Block / Table buttons. |
| `key={topicSlug}` on New Article page was YAGNI | ✅ fixed | Dropped. `key={article.id}` on the Edit page is kept (load-bearing). |
| `.screen-reader-only` legacy `clip: rect()` | ✅ fixed | Added `clip-path: inset(50%)` alongside the legacy rule. |
| `bubbles: true` on `MARKDOWN_TEXTAREA_UPDATE_EVENT` had no consumer or doc | ✅ fixed | One-line comment added explaining it's intentional for future observability/devtools. |
| Space key not intercepted on radios | ✅ fixed | Added to the key list; `preventDefault()` runs; the test at line 123 confirms Space→preventDefault. |

### New findings (nits — all optional)

1. **Test assertion at `markdown-preview-tabs.test.ts:163` is fragile.** `view.getByText("alpha").parentElement?.textContent?.includes("beta")` relies on testing-library matching text nodes + default whitespace normalization + react-markdown's exact `<p>alpha <strong>beta</strong></p>` output. Works today, but intent is unclear. Suggest `assert.match(document.getElementById("content-preview")?.textContent ?? "", /alpha\s+beta/);` for an intent-revealing content check.
2. **`requestAnimationFrame` mock runs synchronously** (`markdown-preview-tabs.test.ts:46`). Test 4 still asserts the correct post-condition (because `updateMarkdownTextarea` set the DOM selection synchronously at line 31 *before* dispatching), so the sync rAF doesn't break the test — but a one-line comment in the mock explaining "sync rAF is intentional because the listener's rAF work is idempotent with the helper's earlier setSelectionRange" would prevent the next reader from "fixing" it.
3. **Split→write branch (empty content on resize) is not covered** — Test 5 only exercises split→preview. The empty-content branch (`return trimmedContentRef.current ? "preview" : "write"`) is one line away. Easy add: render with `defaultValue: ""`, click Split, `matchMediaController.setMatches(false)`, assert Write is checked.
4. **`afterEach` doesn't restore globals** (`markdown-preview-tabs.test.ts:89`). `beforeEach` overwrites globalThis.window/document/etc. each test, so within-file is harmless; Node's default `--test-isolation=process` keeps it from leaking cross-file. A one-line comment in `installDom` documenting the implicit "beforeEach full overwrite" assumption would future-proof against a config change.
5. **`setContent` is recreated each render** (`MarkdownPreviewTabs.tsx:70`). The textarea's `onChange` gets a new function reference each keystroke. Negligible cost. Optional `useCallback` if you want to be tidy.
6. **All three radios' `aria-controls` resolve to their own containing ancestor** (`MarkdownPreviewTabs.tsx:171`). Unconventional but valid — the radios toggle the shell's `data-mode` attribute. Optional: point at `.markdown-editor-grid` instead (the actual child whose layout changes), keeping `aria-controls` inside the radiogroup's children rather than reaching up to the parent. Both shapes work; current shape is honest, just unusual.

### What this commit does particularly well

- `setContent` wrapper is the minimal correct shape — ref-then-state ordering is safe under batching/StrictMode/concurrent rendering, and it unifies the two code paths (`onChange` and the custom-event listener) through one update function. No need to remember to update the ref anywhere else.
- Test scaffolding is exemplary for an authoring surface: per-test fresh `happy-dom` Window, exported `matchMediaController` for behavioral viewport assertions, `IS_REACT_ACT_ENVIRONMENT: true` set, rAF bound to the per-test window. Future tests in this repo can copy the pattern.

### Bottom line

All strong suggestions and nits from the prior review are addressed with surgical, defensible fixes. The PR is ship-ready. The 6 remaining items are polish — pick whichever appeal; ship without if you'd rather move on.

SweetSophia:
## Rereview — PR #233 (commit `207129c`)

**Verdict:** Ready to merge. The single change in this commit is correct, and it walks back a bad call from my prior review.

### The revert is right; my prior YAGNI was wrong

I previously flagged `key={topicSlug}` on the New Article page as YAGNI and you removed it in `e1aad1a`. You restored it in `207129c`. **You were right and I was wrong.**

My error: I reasoned that "the content is always empty so the state is identical." That confuses *initial* state (empty on mount) with *runtime* state. The user can type content into the editor. Under Next.js App Router client-side navigation between two `/wiki/[topicSlug]/new` routes for different topics (Link, browser back/forward, URL bar), the route segment shape is identical — so React reconciles `<MarkdownPreviewTabs>` by tree position and **reuses the instance**, preserving `content`, `activeMode`, and `supportsSplitMode` across the topic change.

Concrete bug without the key:
1. User opens `/wiki/topic-a/new`, types half an article.
2. User navigates to `/wiki/topic-b/new` (e.g., via a bookmark, browser back, or any in-app Link targeting that route).
3. Editor still shows topic-a's content (state reused).
4. User clicks **Create Article** → server action binds to `topic-b` → article created in topic-b with topic-a's body. Silent cross-topic data leak.

`key={topicSlug}` forces a clean unmount/remount on topic change — same defensive pattern as `key={article.id}` on the edit page for the article-reset case. It's load-bearing for App Router client-side nav, not YAGNI. Apologies for the bad call.

### Other 5 follow-up nits

Left unaddressed (all genuinely optional — I flagged them as such, and "ship without" was always a valid choice):
1. Fragile `getByText("alpha")...` assertion in the toolbar-sync test
2. Sync `requestAnimationFrame` mock (no explanatory comment)
3. Split→write branch (empty content on resize) not covered by a test
4. `afterEach` doesn't restore globals (mitigated by Node's default `--test-isolation=process`)
5. `setContent` recreated each render (negligible perf)

All remain valid observations for a future polish pass; none block this PR.

### Bottom line

CI all green on `207129c` (lint/typecheck, CodeQL, Docker, Snyk, tests). Mergeable. The PR has now been through three review cycles and is in good shape — controlled-textarea discipline is solid, ARIA radiogroup is correct, the custom-event protocol is clean, render tests cover the load-bearing paths, and the `key` strategies on both pages are now correctly justified (article.id for edit, topicSlug for new). **Ship it.**

PR #235, opened 2026-06-30
fix(memory): raise recall rate limit

SweetSophia:
## Summary

This pull request (1.10.3) raises the per-client-IP rate limit for the read-only `POST /api/memory/recall` endpoint from **10 to 120 requests per minute** and makes it operator-tunable via a new environment variable, without affecting auth or write-route throttles.

### `src/app/api/memory/recall/route.ts`
- Replaces the hardcoded `{ windowMs: 60_000, maxRequests: 10, keyPrefix: "memory-recall" }` literal with a new `getMemoryRecallRateLimitOptions()` helper.
- Exports `DEFAULT_MEMORY_RECALL_RATE_LIMIT_PER_MINUTE = 120` so the default is part of the module's public surface and directly testable.
- Introduces `readPositiveIntegerEnv(name, fallback)` which:
  - returns the fallback when the env var is unset,
  - returns the fallback when the value doesn't match `/^\d+$/` (e.g. `"240abc"`, `"not-a-number"`, `"1.5"`, negative, zero, decimals),
  - returns the fallback when `Number(value)` isn't a safe positive integer.
- Keeps the existing `keyPrefix: "memory-recall"` so per-client-IP keying and existing rate-limit buckets behave identically — only the per-minute capacity changes.

### New test file: `src/__tests__/api/recall-rate-limit-config.test.ts`
Three regression tests covering the recall rate-limit configuration:
1. **Default behavior** — with the env var unset, `getMemoryRecallRateLimitOptions()` returns the new default (120) and the helper exposes `DEFAULT_MEMORY_RECALL_RATE_LIMIT_PER_MINUTE === 120`.
2. **Valid override** — setting `NOOSPHERE_MEMORY_RECALL_RATE_LIMIT_PER_MINUTE=240` propagates to `maxRequests`.
3. **Invalid override fallback** — for `"0"`, `"-1"`, `"1.5"`, `"240abc"`, `"not-a-number"`, the helper falls back to the default rather than throwing or silently accepting bad input.
Tests clean up the env var in `finally` blocks so they don't leak state into other suites.

### Configuration & docs
- **`noosphere.env.example`**: documents the new `NOOSPHERE_MEMORY_RECALL_RATE_LIMIT_PER_MINUTE=120` knob with rationale ("sized for several local coding CLIs doing prompt-time recall through one shared IP").
- **`README.md`**: adds a paragraph in the endpoint/rate-limit section noting the new default and env override for `POST /api/memory/recall`, alongside the existing write-route body-size notes.
- **`CHANGELOG.md`**: new `[1.10.3] - 2026-06-30` entry under `[Unreleased]` describing this as a patch for "concurrent agent recall reliability" — no schema changes, no breaking changes.

### Versioning
Bumped to **1.10.3** (patch) in:
- `VERSION`
- `docker-compose.yml` (`NOOSPHERE_VERSION` build arg default and `org.opencontainers.image.version` label)
- `hermes-noosphere-memory/plugins/memory/noosphere/plugin.yaml`

## Why
Multiple local coding CLIs auto-recalling memories at prompt time share a single client IP, and the previous 10/minute cap caused legitimate, read-only recall bursts to trip the throttle. Raising the limit and exposing it as an env var preserves protection against abuse while letting operators tune the read-only capacity independently of write-route limits.

## Risk / compatibility
- Behavior change: real-world recall throughput per IP increases ~12× by default. Existing rate-limit keys (`memory-recall:*`) are unchanged, so previously allowed requests remain allowed; only the cap moves.
- No DB schema change, no API contract change (responses and payloads are identical), no auth change.
- Invalid env values silently fall back to the default rather than crashing startup or producing `NaN`/`0` limits.

---

## Summary

This PR raises the default memory recall rate limit to better support concurrent CLI usage, and propagates the new configurable environment variable through all deployment artifacts.

### Changes

**Default rate limit raised to 120 requests/minute** (from the prior lower default) to better accommodate concurrent CLI read workloads.

**Configuration wiring** for `NOOSPHERE_MEMORY_RECALL_RATE_LIMIT_PER_MINUTE` (default: `120`) added to:
- `docker-compose.yml` (app service)
- `docker-compose.noosphere.yml` (both app and bootstrap services)
- `install-openclaw.sh` (default value, env file generation, `export` list, and generated compose template for both services)

**Test refactor** in `src/__tests__/api/recall-rate-limit-config.test.ts`:
- Replaced duplicated try/finally env-var restore blocks with a shared `withRestoredRateLimitEnv` helper for cleaner test setup/teardown.
- Extended invalid-value coverage to include `9007199254740992` (greater than `Number.MAX_SAFE_INTEGER`), guarding against unsafe integer parsing.

---

## Summary

This pull request raises the memory recall API rate limit from a CLI-friendly default to a higher capacity and improves diagnostics around invalid environment overrides.

### Changes

- **Increased default rate limit** for memory recall endpoints to `DEFAULT_MEMORY_RECALL_RATE_LIMIT_PER_MINUTE = 120` requests per minute, up from the previous lower threshold, providing more headroom for concurrent usage.
- **Exported the rate-limit environment variable name** (`MEMORY_RECALL_RATE_LIMIT_ENV`) so tests and external configuration can reference the canonical key without hardcoding the string.
- **Added warn-once diagnostics** for invalid `NOOSPHERE_MEMORY_RECALL_RATE_LIMIT_PER_MINUTE` overrides. The change ensures operators are notified (via `console.warn`) the first time a malformed or out-of-range value is encountered, while still falling back to the default. Duplicate warnings for the same `name=value` pair are suppressed.
- **Updated test suite** (`recall-rate-limit-config.test.ts`) to:
  - Reference the exported `MEMORY_RECALL_RATE_LIMIT_ENV` constant instead of hardcoding the environment variable name.
  - Capture and assert that a warning is emitted for each invalid override value.
  - Handle the dynamic import of the variable name correctly inside the `withRestoredRateLimitEnv` helper.
- **Changelog entry** added under `[1.10.2]` documenting the injected-memory persistence hardening, the restored strict API CI gate, and dependency lockfile refreshes that accompany this patch release.


SweetSophia:
## In-depth review — PR #235 `fix(memory): raise recall rate limit` (1.10.2 → 1.10.3)

**Recommendation: approve.** No correctness defect, no auth bypass, no DoS amplification, no data-integrity risk. The headline change (10 → 120 req/min on the read-only `POST /api/memory/recall`) is the right call for the stated use case, and the env-tunability is narrowly scoped to that one bucket — write/auth limits are deliberately untouched. The findings below are mostly NICE-to-have polish; `M-1` is the one worth fitting in if you can.

I reviewed the route, the test, the shared limiter (`src/lib/rate-limit.ts`), sibling route patterns, and the existing env-parsing utility (`src/lib/obsidian-sync/config.ts:26`), and traced the deployment wiring across `docker-compose.yml`, `docker-compose.noosphere.yml` (app + bootstrap), `install-openclaw.sh` (default + env file + export list + generated compose template), `.env.example`, `noosphere.env.example`, README, CHANGELOG, VERSION, and the four plugin `pkg.json`/`package-lock.json` files. Findings are reconciled against the latest PR head (commit `02953f7`).

---

### Strengths — these should not get refactored by future reviewers

1. **Rate-limit-before-auth ordering is correct for this endpoint.** `requirePermission` ultimately runs a DB `SELECT` on `api_keys` plus an optional NextAuth session lookup — neither free. Rate-limiting first sheds unauthenticated floods cheaply before they hit auth. All 27 rate-limited routes in the repo follow this order; the 12× raise on this *one* read endpoint doesn't change that calculus. (Principle: **defense in depth** — cheap outer ring, expensive inner ring.)
2. **`readPositiveIntegerEnv` validation is correct.** I walked the corner cases manually: `""`, `"  "`, `"+5"`, `"-1"`, `"1.5"`, `"1e3"`, `"0x10"`, `"240abc"`, `"."`, `"9007199254740992"` all correctly fall back via the regex; `"1"`, `"120"`, `"9007199254740991"` (`MAX_SAFE_INTEGER`) correctly parse; `"0"` and `"00"` fall back via `> 0`. Nothing that should fall back slips through; nothing that should parse is rejected. The combination `/^\d+$/` + `Number.isSafeInteger` + `> 0` is the canonical pattern for this requirement. (Principle: **fail-safe defaults** — bad input silently returns the safe default rather than crashing startup or producing `NaN`/`0` limits.)
3. **Test mutation ordering is correct.** The tests mutate `process.env` *before* the dynamic `await import(...)`. Because `getMemoryRecallRateLimitOptions()` reads env at call time (not module-load time), even a cached module sees the fresh value. This is the right design and exploited correctly.
4. **Patch bump (1.10.2 → 1.10.3) is appropriate.** A default-relaxing throttle change is backwards-compatible; downstream agents that previously got 429s and built retry logic will just see less throttling. CHANGELOG is explicit about the change and the env knob. Minor would have been defensible (12× is substantial), but patch matches the "fix a previously over-strict default" framing.
5. **Narrow scope is the right policy.** I counted: `memory-save` 30, `memory-get` 30, `memory-status` 60, `memory-settings-post` 30, `articles-post` 30, `articles-patch` 30, `articles-delete` 30. None of the write/settings routes are touched. 120 now sits 2× above the next-highest bucket (60 on `memory-status` / `articles-get`), which is reasonable for a read-only recall endpoint hit by multiple CLIs. CHANGELOG explicitly notes "without weakening auth or write-route throttles." (Principle: **separation of concerns** + **least privilege** — one tunable knob, in the right place.)
6. **`keyPrefix: "memory-recall"` is unchanged**, so deployments don't reset anyone's existing Redis ZSET bucket — they just start honoring the higher cap on the next request. Smooth migration.
7. **Deployment wiring is comprehensive.** Every surface that ships a `NOOSPHERE_MEMORY_RECALL_RATE_LIMIT_PER_MINUTE` default of 120 is touched: `docker-compose.yml`, `docker-compose.noosphere.yml` (both `app` and `bootstrap` services), `install-openclaw.sh` (default value, generated `.env` line, `export` list, and the inline generated compose template for both services), `.env.example`, `noosphere.env.example`. Plus the version bump is propagated to `VERSION`, `pkg.json`, four plugin `pkg.json` + `package-lock.json` pairs, the hermes `plugin.yaml`, and the `NOOSPHERE_VERSION` build arg + `org.opencontainers.image.version` label. No missed surface.

---

### Findings

#### M-1 — Missing `Retry-After` header on 429 (Medium)

**File:** `src/lib/rate-limit.ts:54` — pre-existing across all 27 rate-limited routes, but this is the right PR to fix it.

```ts
if (count >= options.maxRequests) {
  return {
    allowed: false,
    response: apiError("Too many requests", 429),
  };
}
```

`apiError` returns a vanilla `NextResponse.json({ error, requestId }, { status })` with no `Retry-After`. For an AI/agent-facing endpoint where clients (per the PR's own motivation) generate "prompt-time auto-recall bursts," the absence of `Retry-After` forces every client to invent its own backoff and get it wrong. RFC 6585 and RFC 7231 §7.1.3 both call out `Retry-After` as the canonical signal on 429.

The PR's motivation is exactly the scenario where this matters most: legitimate bursts that *almost* hit the cap need a clean way to back off rather than hammer-and-hope. Three-line fix in `rate-limit.ts`:

```ts
const response = apiError("Too many requests", 429);
response.headers.set("Retry-After", String(Math.ceil(options.windowMs / 1000)));
return { allowed: false, response };
```

For a sliding-window limiter, setting `Retry-After` to one full window (60s) is the canonical worst-case wait. The worst-case client behavior — fixed 60s sleep — is exactly correct.

#### M-2 — `readPositiveIntegerEnv` duplicates an existing env-parsing utility (Medium)

**Files:** new `src/app/api/memory/recall/route.ts:58-66` vs. existing `src/lib/obsidian-sync/config.ts:26-30`.

There is already a `parseEnvInt(env, fallback)` helper in `obsidian-sync/config.ts`:

```ts
function parseEnvInt(env: string | undefined, fallback: number): number {
  if (env === undefined) return fallback;
  const n = parseInt(env, 10);
  return isNaN(n) ? fallback : n;
}
```

It is permissive — `parseInt("1abc", 10) === 1`, accepts negatives, doesn't bound to a safe integer, has no warn-once — so `readPositiveIntegerEnv` is *not* a drop-in replacement. But having two env-int parsers with different semantics in the same repo is the exact consistency drift that makes future env knobs harder to harden correctly. This is the natural moment to promote the stricter helper to `src/lib/env-config.ts` (e.g. `parsePositiveIntEnv(name, fallback)`) and tighten `parseEnvInt` to delegate, so the next env knob inherits the safer semantics for free. Minor refactor; non-blocking.

#### M-3 — CHANGELOG skips the [1.10.2] entry (Medium, doc gap)

**File:** `CHANGELOG.md` — the file jumps `[Unreleased]` → `[1.10.3]` with no `[1.10.2]` entry in between, even though `VERSION` goes `1.10.2 → 1.10.3` and the lockfiles / plugin `pkg.json` files were all bumped to 1.10.2 in a prior release. This means the CHANGELOG is currently out of sync with reality: either 1.10.2 was shipped without a CHANGELOG entry (a process gap), or the bump in this PR skipped backfilling it. Either way, the CHANGELOG as submitted is missing one version. Backfill a `[1.10.2]` entry above `[1.10.3]` (pull the commits between the two tags from `git log v1.10.2..v1.10.3` or `git log 1.10.1..HEAD~1`) so the changelog reflects every shipped release. This is also flagged by the kilo-code-bot review.

#### L-1 — Per-IP keying trusts client-supplied headers (Low, pre-existing, out of scope)

`src/lib/rate-limit.ts:11-18` reads `x-real-ip` → `cf-connecting-ip` → `x-forwarded-for` → `"unknown"`. If the app is ever deployed without a sanitizing front-proxy that strips and re-sets these headers, an attacker can vary them per request and bypass the per-IP limit entirely. Pre-existing across all 27 routes; the 12× raise on this one read endpoint is well within the honest-client / shared-egress case the CHANGELOG describes. Not introduced or worsened by PR #235. Worth filing as a separate hardening ticket (e.g. introduce `TRUSTED_PROXY_CIDR` and document the assumption in `README.md` / `noosphere.env.example`).

#### L-2 — Sliding-window atomicity gap when `pipeline.exec()` returns `null` (Low, pre-existing)

`src/lib/rate-limit.ts:42-45` treats a `null` pipeline result as fail-open. Correct for total Redis loss (and documented as such), but this path is also taken if a single command in the pipeline individually errors and ioredis returns `null`. Pre-existing; not in scope for this PR. Flagging because the rate-limit-first design makes the fail-open behavior a meaningful DoS lever during partial Redis failures.

#### L-3 — Bucket migration on deploy is undocumented (Low)

The CHANGELOG entry describes the new default and the env knob but doesn't mention that pre-existing Redis buckets carry over. Because `keyPrefix: "memory-recall"` is unchanged, an IP that was at 9/10 under 1.10.2 can immediately post-deploy submit ~112 more requests before the stale entries age out (60s window), getting ~121 requests in that minute. This is correct sliding-window behavior and what the operator wants ("raise the cap, don't reset state"), but a one-line note in the CHANGELOG ("existing rate-limit buckets are preserved across the upgrade; the cap simply moves up") would set operator expectations. (Principle: **least astonishment** — same-key-prefix behavior should be visible.)

#### N-1 — Silent env fallback hides operator typos (Nit)

`readPositiveIntegerEnv` returns the default for any malformed value (zero, negative, decimal, non-numeric, unsafe integer) with no log. An operator who sets `NOOSPHERE_MEMORY_RECALL_RATE_LIMIT_PER_MINUTE=120abc` to "tune" the limit gets the default with zero indication their setting was ignored — the classic 3-AM postmortem setup. A one-shot warn keyed by `(name, value)` (so a misconfigured value doesn't spam the log at request frequency) would surface typos without log-flooding. The kilo-code-bot review independently flagged this same concern. Cheap to add; purely additive.

#### N-2 — Leading zeros silently accepted (Nit)

`/^\d+$/` matches `"0000120"` → 120. A stricter `/^(0|[1-9]\d*)$/` would reject leading zeros and catch operator typos or padded values from config-management tooling. Not a security issue (`Number("0000120") === 120`). Stylistic.

#### N-3 — Test does not explicitly cover the empty-string case (Nit)

The default test does `delete process.env[NAME]` and checks the fallback. It does not assert that an *explicitly set* `""` also falls back to the default. Your code *does* handle it correctly (`!/^\d+$/.test("")` → fallback), and the test's invalid set doesn't include `""`. One-line addition if you want to lock the regex behavior. Tiny coverage gap.

#### N-4 — `MEMORY_RECALL_RATE_LIMIT_ENV` is module-private but conceptually a public contract (Nit)

It is declared `const` (not `export`), which is the right call for module hygiene — the env name doesn't need to be public. But note that the test file currently hardcodes `"NOOSPHERE_MEMORY_RECALL_RATE_LIMIT_PER_MINUTE"` as a string literal rather than reading it from the module, so the literal is duplicated across `route.ts` and the test. If the env name ever changes, the test silently keeps testing the old literal. Trade-off is fine as-is; just be aware.

---

### Notes I considered and chose NOT to flag

- **`getMemoryRecallRateLimitOptions()` is called per request and re-reads `process.env` each time.** Trivial; `process.env` access is fast. Memoizing would buy nothing (env doesn't change without a restart) and would only obscure the live-read intent that the tests rely on. Leave as-is.
- **All sibling routes use inline rate-limit option literals; this PR introduces a helper function for the recall route only.** This is a deliberate break from pattern justified by env-tunability. If a second route becomes env-tunable, lift the pattern to a shared helper in `src/lib/rate-limit.ts` (e.g. `buildRateLimitOptions(prefix, envName, fallback)`) and convert the other routes later. Not blocking.
- **Pre-existing `429` body shape (`{ error, requestId }`).** Out of scope.

---

### OWASP API Security Top 10 (2023) trace

- **API4:2023 Unrestricted Resource Consumption** — *addressed* by this PR (relaxed default is intentional and narrow; knob is operator-tunable; fail-open is hard-only under total Redis loss). Tightened further by M-1 (`Retry-After`).
- **API1:2023 Broken Object Level Authorization** — *not touched*; recall's RBAC lives in `executeMemoryRecallRequest` via `allowedScopes`, unchanged.
- **API2:2023 Broken Authentication** — *not weakened*; auth still happens after the throttle and unauthenticated bursts are still shed at the limiter.
- **API3:2023 Broken Object Property Level Authorization** — n/a.

---

### Semantic anchors / principles invoked

- **Fail-safe defaults** ✓ (`readPositiveIntegerEnv` falls back, doesn't crash)
- **Defense in depth** ✓ (rate-limit-first); *dented* by missing `Retry-After` (M-1)
- **Least astonishment** ✓ (same keyPrefix, smooth migration); partially undocumented (L-3)
- **Separation of concerns** ✓ (only the recall bucket is tunable)
- **Principle of least privilege** ✓ (only the read-only recall path is widened)
- **DRY** — *dented* by M-2 (two env-int parsers)

---

**TL;DR:** Ship it after backfilling the missing `[1.10.2]` CHANGELOG entry (M-3 — a one-time doc fix). Two small polish opportunities (`Retry-After` header, lift the env parser to `src/lib/env-config.ts`) and one tiny silent-fallback warn (N-1). Everything else is nit-or-better. The deployment-wiring breadth and the test design are notably thorough.

PR #236, opened 2026-07-02
test: cover auth proxy rate limiting

SweetSophia:
## Summary
Add proxy-level test coverage for the existing authentication rate limiters on the NextAuth credentials callback and the wiki login page, verifying that the 10 requests/minute/IP threshold permits the 10th request and rejects the 11th with HTTP 429.

## Changes
- **`src/__tests__/security/proxy.test.ts`**: Added two new test cases alongside the existing article mutation rate-limit test.
  - **"proxy rate-limits NextAuth callback attempts"** — Sends repeated `POST` requests to `/api/auth/callback/credentials` using a unique `x-real-ip` header. After 10 requests the response is confirmed not to be 429, and the 11th request is asserted to return 429.
  - **"proxy rate-limits the wiki login page"** — Sends repeated `GET` requests to `/wiki/login` using a unique `x-real-ip` header, with the same threshold check (10 permitted, 11th blocked with 429).

## Technical Notes
- Each test uses a unique `x-real-ip` value derived from `crypto.randomUUID()` to avoid cross-test interference in the limiter's client-key state.
- The test for the credentials callback uses `POST` to match the actual auth submission flow, while the wiki login test uses `GET` since that is how the login page is served.
- These tests exercise the `auth` rate limiter already applied by the proxy to `/api/auth*` and `/wiki/login`; no production code changes are required.
- The threshold semantics (10 allowed, 11th blocked) are explicitly validated by separating the final blocking request from the loop that performs the first 10 requests.

## Impact
- Locks in the current rate-limiting contract for sensitive authentication surfaces, ensuring regressions that disable or misconfigure the proxy's `auth` limiter (on either the credentials callback or the login page) will be caught by `npm run test:security`.
- No runtime behavior change; production rate-limit configuration is unchanged.

---

Add tests covering rate limiting for authentication-related proxy routes.

- Extract the rate-limit assertion logic from the existing article mutation test into a reusable `expectRateLimitedAfter` helper that drives requests under a unique client IP until a 429 is returned.
- Refactor the article PATCH test to use the new helper, verifying the 30-request threshold.
- Add a test confirming `/api/auth/callback/credentials` returns 429 after 10 POSTs from the same IP.
- Add a test confirming `/wiki/login` returns 429 after 10 GETs from the same IP.

---

Adds test coverage for auth proxy rate limiting

---

## Summary

This PR expands the existing proxy security test suite to provide comprehensive coverage of the rate-limiting behavior on authentication-related routes. Previously, only the article mutation route (`/api/articles/article-1`) had dedicated rate-limit coverage; the new tests formalize and verify the auth proxy's behavior across multiple entry points.

## Key Changes

**Introduced a reusable test helper** (`expectRateLimitedAfter`) that:
- Sends `maxRequests` requests with a unique `x-real-ip` header and asserts each returns `200` (forwarded to the downstream handler)
- Sends one additional request and asserts it returns `429` (rate-limited by the proxy)

This helper replaces the previous hand-rolled loop in the article-mutation test and is now reused across multiple scenarios.

**New test cases added:**

1. **NextAuth callback rate limiting** — Verifies that `POST /api/auth/callback/credentials` returns `429` after 10 requests from the same IP.
2. **NextAuth non-callback route rate limiting** — Verifies the same threshold (10 requests) applies to a representative non-callback route such as `/api/auth/signin`.
3. **Wiki login page rate limiting** — Verifies that `GET /wiki/login` returns `429` after 10 requests from the same IP.
4. **Per-IP isolation** — Confirms that saturating one IP's budget does not block requests from a different IP against the wiki login route.
5. **Shared auth budget across entry points** — Verifies that the wiki login page and NextAuth callback route draw from the same rate-limit bucket for a given IP: 5 login requests + 5 callback POSTs exhaust the 10-request budget, causing the 11th login request to return `429`.

**Refactored existing test:** The article-mutation rate-limit test now uses the shared helper instead of inlining its loop, making the expected threshold (30 requests) explicit in the test signature.

## Impact

These tests serve as executable documentation of the auth proxy's rate-limiting contract — confirming that login pages and NextAuth endpoints share a unified per-IP budget, that the threshold is enforced, and that limits are properly isolated across distinct client IPs. They guard against regressions in the rate-limiting middleware configuration.

### DIFF
--- a/src/__tests__/security/proxy.test.ts
+++ b/src/__tests__/security/proxy.test.ts
@@ -26,6 +26,29 @@ function request(path: string, init: RequestInit = {}) {
   });
 }
 
+async function expectRateLimitedAfter({
+  path,
+  method = "GET",
+  maxRequests,
+  ipPrefix,
+}: {
+  path: string;
+  method?: string;
+  maxRequests: number;
+  ipPrefix: string;
+}) {
+  const headers = { "x-real-ip": `${ipPrefix}-${crypto.randomUUID()}` };
+
+  for (let i = 0; i < maxRequests; i += 1) {
+    const response = await proxy(request(path, { method, headers }));
+    // 200 here means the proxy let the request continue to the downstream handler.
+    assert.equal(response.status, 200);
+  }
+
+  const blocked = await proxy(request(path, { method, headers }));
+  assert.equal(blocked.status, 429);
+}
+
 test("proxy adds security headers to wiki pages", async () => {
   const response = await proxy(request("/wiki/projects/example"));
 
@@ -36,16 +59,72 @@ test("proxy adds security headers to wiki pages", async () => {
 });
 
 test("proxy rate-limits article mutation routes", async () => {
-  const headers = { "x-real-ip": `patch-${crypto.randomUUID()}` };
-  let response = await proxy(request("/api/articles/article-1", { method: "PATCH", headers }));
+  await expectRateLimitedAfter({
+    path: "/api/articles/article-1",
+    method: "PATCH",
+    maxRequests: 30,
+    ipPrefix: "patch",
+  });
+});
 
-  for (let i = 1; i < 30; i += 1) {
-    response = await proxy(request("/api/articles/article-1", { method: "PATCH", headers }));
+test("proxy rate-limits NextAuth callback attempts", async () => {
+  await expectRateLimitedAfter({
+    path: "/api/auth/callback/credentials",
+    method: "POST",
+    maxRequests: 10,
+    ipPrefix: "auth",
+  });
+});
+
+test("proxy rate-limits a representative NextAuth non-callback route", async () => {
+  await expectRateLimitedAfter({
+    path: "/api/auth/signin",
+    maxRequests: 10,
+    ipPrefix: "auth-signin",
+  });
+});
+
+test("proxy rate-limits the wiki login page", async () => {
+  await expectRateLimitedAfter({
+    path: "/wiki/login",
+    maxRequests: 10,
+    ipPrefix: "login",
+  });
+});
+
+test("proxy auth rate limits are isolated per IP", async () => {
+  const path = "/wiki/login";
+  const saturatedIpHeaders = { "x-real-ip": `auth-ip-a-${crypto.randomUUID()}` };
+  const otherIpHeaders = { "x-real-ip": `auth-ip-b-${crypto.randomUUID()}` };
+
+  for (let i = 0; i < 10; i += 1) {
+    const response = await proxy(request(path, { headers: saturatedIpHeaders }));
+    assert.equal(response.status, 200);
   }
 
-  assert.notEqual(response.status, 429);
+  const blocked = await proxy(request(path, { headers: saturatedIpHeaders }));
+  assert.equal(blocked.status, 429);
 
-  const blocked = await proxy(request("/api/articles/article-1", { method: "PATCH", headers }));
+  const otherIpResponse = await proxy(request(path, { headers: otherIpHeaders }));
+  assert.equal(otherIpResponse.status, 200);
+});
+
+test("proxy shares the auth rate-limit budget across login and auth API routes", async () => {
+  const headers = { "x-real-ip": `auth-shared-${crypto.randomUUID()}` };
+
+  for (let i = 0; i < 5; i += 1) {
+    const response = await proxy(request("/wiki/login", { headers }));
+    assert.equal(response.status, 200);
+  }
+
+  for (let i = 0; i < 5; i += 1) {
+    const response = await proxy(
+      request("/api/auth/callback/credentials", { method: "POST", headers })
+    );
+    assert.equal(response.status, 200);
+  }
+
+  const blocked = await proxy(request("/wiki/login", { headers }));
   assert.equal(blocked.status, 429);
 });
 


SweetSophia:
## In-depth review — PR #236 `test: cover auth proxy rate limiting`

**Verdict: approve. No errors.** The two new tests are meaningful (not vacuous), the helper extraction is a clean improvement over the reviewed commit, and the boundary semantics are exact. CI is green and a local `npm run test:security` re-run confirms all 17 tests pass. Findings below are coverage enhancements and small polish — nothing blocks merge.

### What's correct — and worth saying so

- **The tests are not vacuous.** I traced the full chain: for paths that pass the proxy's rate-limit gate, `proxy()` returns `addSecurityHeaders(NextResponse.next())` (`src/proxy.ts:132`). `NextResponse.next()` is a status-200 response. So `assert.equal(response.status, 200)` asserts a real proxy behavior (request was allowed through, not blocked), not a stub.
- **`FakeRedisClient` faithfully emulates the sliding-window ZSET.** `pipeline.exec()` runs `zremrangebyscore → zcard → zadd → expire` synchronously in array.map order, so `zcard` observes the pre-`zadd` count. `zremrangebyscore` drops everything in `[-inf, windowStart]`. Member uniqueness matches real ZSET NX. Result shape `[null, number]` matches the `[[err, res], ...]` format `rate-limit.ts:48-49` reads.
- **Boundary semantics are correct.** `rate-limit.ts:51` checks `count >= maxRequests` where `count` is the pre-`zadd` `zcard`. So for `maxRequests: 10`, requests 1–10 pass (zcard 0→9, all `< 10`), request 11 is blocked (zcard=10, `10 >= 10`). The helper's `for (i < maxRequests) { assert 200 }` then one final `assert 429` correctly expresses "maxRequests allowed, (maxRequests+1)th blocked" for both the 10/min auth case and the 30/min api-write case. The off-by-one is clean.
- **Test isolation is robust.** Each test gets a fresh `FakeRedisClient` via `_redisTestHooks.setClientForTesting(new FakeRedisClient() as never)` in `beforeEach`, plus a per-test `${ipPrefix}-${crypto.randomUUID()}` `x-real-ip`. No state leaks, no ordering dependency.
- **The three tests cover distinct proxy gate paths** (I traced each):
  - `/api/auth/callback/credentials` POST → `path.startsWith("/api/auth")` → `keyPrefix: "auth"`, `maxRequests: 10` ✓
  - `/wiki/login` GET → `path === "/wiki/login"` → `keyPrefix: "auth"`, `maxRequests: 10` ✓
  - `/api/articles/article-1` PATCH → `isWriteMethod && isWriteEndpoint` → `keyPrefix: "api-write"`, `maxRequests: 30` ✓
- **The force-push from `3e23a6d5` → `8c9fb2e` is a clean response to earlier feedback.** Sourcery asked for a helper and stricter assertions; Gemini asked for a single loop. Both addressed: `expectRateLimitedAfter` extracts the shared pattern, and `assert.notEqual(429)` becomes `assert.equal(200)` (stricter, more precise — see Finding #6 below for the trade-off). The old "1 initial + loop(9)" → new "loop(10)" conversion preserved the exact request counts (10/30), so the boundary semantics are identical to what was reviewed.

---

### Findings

#### M-1 — Per-IP isolation not tested at the proxy layer (Medium, coverage gap)

The PR proves one IP exhausts its bucket. It does NOT prove a different IP is unaffected when one IP is blocked. The per-IP keying (`x-real-ip → getClientIdentifier → ratelimit:${keyPrefix}:${clientId}`) is exercised at the lib level in `src/__tests__/rate-limit.test.ts:49` (`"uses unique keys per client IP"`), but if the proxy ever lost the `x-real-ip` header (a `headers.get` regression, a casing bug, a middleware-chain stripping it, or a future move to `cf-connecting-ip`/`x-forwarded-for`), the lib-level test wouldn't catch it because it constructs `NextRequest` directly. Suggested follow-up:

```ts
test("proxy rate-limits are isolated per IP", async () => {
  const ipA = `ipA-${crypto.randomUUID()}`;
  const ipB = `ipB-${crypto.randomUUID()}`;
  const path = "/wiki/login";

  for (let i = 0; i < 10; i += 1) {
    const r = await proxy(request(path, { headers: { "x-real-ip": ipA } }));
    assert.equal(r.status, 200);
  }
  assert.equal(
    (await proxy(request(path, { headers: { "x-real-ip": ipA } }))).status,
    429,
  );
  assert.equal(
    (await proxy(request(path, { headers: { "x-real-ip": ipB } }))).status,
    200,
  );
});
```

Not blocking — but cheap, high-signal, and locks in the per-IP contract at the proxy level (not just the lib level).

#### M-2 — 429 response has no `Retry-After` header (Medium, cross-ref PR #235 M-1)

This is the same finding I raised on PR #235. `apiError("Too many requests", 429)` returns `NextResponse.json({ error, requestId }, { status: 429 })` with no `Retry-After`. RFC 6585 explicitly calls out `Retry-After` as the canonical signal on 429. The auth and login surfaces are exactly where this matters most — clients hitting the credential callback want a clean back-off signal rather than hammer-and-hope. This PR is a natural second place to fix it (after the rate-limit lib fix in #235): either assert `Retry-After: 60` in the new tests, or add the header in `apiError`/`rate-limit.ts`. Worth tracking as one item across both PRs.

#### L-1 — Other NextAuth routes not covered (Low, coverage gap)

`src/proxy.ts:60` rate-limits the entire `/api/auth*` namespace (`/signin`, `/signout`, `/session`, `/csrf`, `/providers`, …). Only `/api/auth/callback/credentials` is tested. A typo turning `startsWith("/api/auth")` into `=== "/api/auth/callback/credentials"` would still pass the current tests. Adding one representative path (e.g. `/api/auth/signin`) is one line and locks in the wildcard.

#### L-2 — `api-write` rate-limit only tested on `/api/articles/*` (Low, acceptable)

`src/proxy.ts:73-82` lists 9 write endpoints sharing the same `api-write` bucket. Only `/api/articles/article-1` PATCH is tested. A typo dropping a path from `isWriteEndpoint` wouldn't be caught — but it would also break production immediately. Testing each path is noise; skip unless a stronger gate is wanted.

#### L-3 — Cross-endpoint coupling in the shared `auth` bucket is not exercised (Low, coverage gap — flagged independently by kilo-code-bot)

`src/proxy.ts:60-67` uses the SAME `keyPrefix: "auth"` for both `/api/auth*` and `/wiki/login`. The Redis key is `ratelimit:auth:${clientId}` — so a single attacker IP's quota is **shared across both endpoints** (10/min combined, not 10/min per endpoint). This is a deliberate defense-in-depth choice (one IP can't burn the auth surface by spreading across endpoints) and is reasonable for the typical login flow (GET csrf → POST callback is 2 reqs, well under the limit). But the two new tests use different `ipPrefix + randomUUID()` per test, so each test exercises its endpoint with a fresh bucket. **Neither test validates that a single IP's quota is shared across endpoints.** A follow-up test using the SAME IP for both `/api/auth/callback/credentials` and `/wiki/login` (e.g. send 6 of each, expect all 12 to pass if the design were per-endpoint, or expect the 11th-12th to 429 if shared) would lock in the intended behavior. Worth noting: the kilo-code-bot review surfaced this same observation as a bonus finding.

#### L-4 — Window-reset behavior not tested (Low, acceptable)

After 60s, expired entries are dropped by `zremrangebyscore`, unblocking the IP. No test verifies this. Hard to test without injecting a clock into `FakeRedisClient` — invasive and out of scope. Acceptable gap.

#### L-5 — `assert.equal(response.status, 200)` is stricter than the original `notEqual(429)` (Low, defensible)

The OLD article test used `assert.notEqual(response.status, 429)`. The NEW helper uses `assert.equal(response.status, 200)`. Today both pass because `NextResponse.next()` is always 200. The stricter assertion couples the test to the proxy's pass-through status — a future change that, say, redirects unauthenticated PATCH to `/wiki/login` (returning 307) would break this test even though rate-limiting itself still works. **Defensible trade-off: stricter assertions catch more regressions, and `notEqual(429)` was strictly weaker.** Keep the strict version. Worth a one-line comment in the helper clarifying "200 = proxy passed the request through; this test exercises the limiter, not the downstream handler."

#### N-1 — Inconsistency between new helper and the negative test (Nit)

`"proxy does not rate-limit article reads"` (line 94) still uses the old `assert.notEqual(response.status, 429)` inline pattern. Justified because the test asserts *absence* of rate-limiting (and the helper's positive assertion doesn't cover that case), but the codebase now has two assertion styles for the same proxy. A parallel `expectNeverRateLimited({ path, maxRequests })` helper would close the symmetry gap. YAGNI unless a second negative test appears.

#### N-2 — `ipPrefix` is cosmetic (Nit)

`${ipPrefix}-${crypto.randomUUID()}` is functionally equivalent to `${crypto.randomUUID()}`. The prefix only aids log triage (test failures show which test's IP got blocked). Fine as-is; the readability win is real.

#### N-3 — `expectRateLimitedAfter` name (Nit)

"After" reads as "after maxRequests attempts, the next is blocked" — exactly what the helper does. Name is fine. Alternatives (`expectAllowedUntilLimit`, `expectBlockedOnRequest(maxRequests + 1)`) are not clearer. Keep.

#### N-4 — POST without body/CSRF is not a realistic credentials-callback shape (Nit)

A real NextAuth credentials POST carries a CSRF token, `callbackUrl`, and credential fields. The test sends a bare POST. Acceptable — the test exercises the proxy's rate-limit gate, not NextAuth's handler, and NextAuth's behavior is outside the proxy's responsibility. Not a defect.

#### N-5 — `assert.equal` vs `assert.strictEqual` (Nit)

The file imports `assert from "node:assert/strict"`, so `equal` is already strict-by-default. No actual bug. But the rest of the codebase uses `assert.deepStrictEqual` / `assert.strictEqual` explicitly (e.g. `rate-limit.test.ts:29`). For stylistic consistency, `assert.strictEqual` would be more explicit. Minor.

---

### On the question — "did I make errors?"

**No errors found.** The refactor is a clean improvement over the reviewed commit `3e23a6d5` (helper extracted, stricter assertion, single loop — all addressing earlier feedback while preserving the exact boundary counts). The tests are non-vacuous against a faithful FakeRedisClient, the boundary semantics are exact, isolation is robust, and the three tests cover distinct proxy gate paths. The findings above are coverage enhancements and stylistic nits, not defects. Ship it.

---

### Semantic anchors / principles invoked

- **Regression test** ✓ — locks in the rate-limit contract on auth surfaces
- **Boundary value analysis** ✓ — tests exactly at the threshold (Nth allowed, N+1th blocked)
- **Test isolation** ✓ — fresh FakeRedisClient per test + per-test UUID x-real-ip
- **DRY** ✓ — `expectRateLimitedAfter` extracts the shared "send N, assert blocked on N+1" pattern
- **Single Responsibility** ✓ — `proxy.test.ts` tests the proxy layer (limiter engagement), not the downstream app handler
- **Defense in depth** ✓ — the auth limiter protects login/credential endpoints independently of the api-write limiter
- **Non-vacuous assertion** ✓ — the `equal(200)` assertion has real meaning because `NextResponse.next()` is a real proxy behavior, not a stub; FakeRedisClient is a faithful ZSET emulation

---

**TL;DR:** Approve, no errors. Two coverage enhancements (per-IP isolation test, Retry-After header — the latter shared with PR #235), a few stylistic nits. The force-push delta is a clean response to earlier review feedback and preserved the exact boundary semantics. CI green, local test re-run green.

PR #237, opened 2026-07-03
test: cover article update rate and body limits

SweetSophia:
## Summary

This PR adds focused B7 coverage for the article update (`PATCH /api/articles/[id]`) endpoint's security boundaries.

## Changes

- Adds `src/__tests__/security/articles-update.test.ts` coverage proving the route-local article PATCH limiter runs before auth work: the first 30 unauthenticated requests return 401, and the 31st returns 429 from the limiter.
- Extends `src/__tests__/api/bounded-json-routes.integration.test.ts` with a DB-backed integration case proving oversized article PATCH bodies return HTTP 413 from the bounded JSON reader.
- Wires `npm run test:security` into CI so the route-level rate-limit coverage is enforced by the merge gate.
- Removes the redundant source-text parser assertion; the 413 integration test now carries the bounded-parser regression coverage behaviorally.

## Verification

- `npm run test:security` — 21/21 passing
- `npm run test:api` — 350/350 passing against an isolated temporary Postgres 16 database
- `npm run lint` — passes with one pre-existing unrelated warning in `src/app/wiki/admin/topics/actions.ts`
- `npm run typecheck` — passes
- Two sequential review passes on the follow-up diff — no blockers

## Summary

This PR adjusts the security test suite for the PATCH article endpoint and wires security tests into CI:

### CI Workflow Changes (`.github/workflows/ci.yml`)
- Adds a new CI step that runs `npm run test:security`, ensuring the security-focused tests execute automatically on every build alongside the existing API tests.

### Security Test Updates (`src/__tests__/security/articles-update.test.ts`)
- **Removes the static-source JSON parser test**: Drops a test that read the route source file from disk and asserted it used `readBoundedJsonObject` instead of `request.json()`. The test no longer relies on filesystem access or source-string inspection.
- **Updates comments on the rate-limit test**: Clarifies the test's intent — the first 30 unauthenticated requests return 401 (because they lack credentials, without DB work), and the 31st request returning 429 demonstrates that the route-local limiter runs before authentication. The accompanying comment about "30 requests" is now made explicit instead of describing it indirectly.
- **Cleans up unused imports**: `readFile` from `node:fs/promises` and `path` from `node:path` are removed since the deleted test was their only consumer.

### Net Effect
The security test suite for article updates is now slimmer (one fewer test that performed file-system inspection) but better documented, and the remaining tests are now automatically executed in CI through the new `test:security` job step.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -85,3 +85,6 @@ jobs:
 
       - name: Run API tests
         run: npm run test:api
+
+      - name: Run security tests
+        run: npm run test:security
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "test:cache": "node --env-file=.env.test --import tsx --test 'src/__tests__/cache/**/*.test.ts'",
     "test:rate-limit": "node --env-file=.env.test --import tsx --test src/__tests__/rate-limit.test.ts",
     "test:api": "node --import tsx --test 'src/__tests__/api/**/*.test.ts'",
-    "test:security": "node --import tsx --test src/__tests__/security/uploads.test.ts src/__tests__/security/proxy.test.ts",
+    "test:security": "node --env-file=.env.test --import tsx --test src/__tests__/security/uploads.test.ts src/__tests__/security/proxy.test.ts src/__tests__/security/articles-update.test.ts",
     "lint": "eslint",
     "typecheck": "tsc --noEmit",
     "db:migrate": "prisma migrate dev",
--- a/src/__tests__/api/bounded-json-routes.integration.test.ts
+++ b/src/__tests__/api/bounded-json-routes.integration.test.ts
@@ -124,12 +124,74 @@ test("legacy JSON routes enforce size and nesting limits (issue #192)", async ()
       where: { authorName: expectedAuthorName },
       select: { authorName: true, type: true },
     });
-    // Exactly one log is expected here: the empty-body lint success at the start.
-    assert.deepEqual(auditLogs, [{ authorName: expectedAuthorName, type: "lint" }]);
     const deletedLogs = await prisma.activityLog.deleteMany({
       where: { authorName: expectedAuthorName },
     });
+    await prisma.apiKey.delete({ where: { id: apiKey.id } });
+
+    // Exactly one log is expected here: the empty-body lint success at the start.
+    assert.deepEqual(auditLogs, [{ authorName: expectedAuthorName, type: "lint" }]);
     assert.equal(deletedLogs.count, 1);
+  }
+});
+
+test("article PATCH rejects oversized request bodies", async () => {
+  const { prisma } = await import("@/lib/prisma");
+  const { PATCH: updateArticle } = await import("@/app/api/articles/[id]/route");
+  const runId = crypto.randomUUID();
+  const rawKey = `noo_${crypto.randomBytes(32).toString("base64url")}`;
+  const keyHash = crypto.createHash("sha256").update(rawKey).digest("hex");
+
+  const apiKey = await prisma.apiKey.create({
+    data: {
+      name: `${TEST_PREFIX}${runId}`,
+      keyHash,
+      keyPrefix: rawKey.slice(0, 8),
+      permissions: "ADMIN",
+      allowedScopes: ["*"],
+    },
+  });
+
+  try {
+    const topic = await prisma.topic.create({
+      data: {
+        name: `${TEST_PREFIX}${runId}`,
+        slug: `${TEST_PREFIX}${runId}`,
+      },
+    });
+    const article = await prisma.article.create({
+      data: {
+        title: `${TEST_PREFIX}${runId}`,
+        slug: `${TEST_PREFIX}${runId}`,
+        content: "Original content",
+        topicId: topic.id,
+      },
+    });
+    const oversizedArticlePatchBody = JSON.stringify({
+      content: "x".repeat(
+        ARTICLE_LIMITS.maxContentSize + DEFAULT_JSON_BODY_MAX_BYTES + 1,
+      ),
+    });
+    const response = await updateArticle(
+      buildRequest(
+        rawKey,
+        `/api/articles/${article.id}`,
+        oversizedArticlePatchBody,
+        "PATCH",
+      ),
+      { params: Promise.resolve({ id: article.id }) },
+    );
+    assert.equal(response.status, 413);
+  } finally {
+    await prisma.activityLog.deleteMany({
+      where: { authorName: `${TEST_PREFIX}${runId}` },
+    });
+    await prisma.article.deleteMany({
+      where: { slug: `${TEST_PREFIX}${runId}` },
+    });
+    await prisma.topic.deleteMany({
+      where: { slug: `${TEST_PREFIX}${runId}` },
+    });
     await prisma.apiKey.delete({ where: { id: apiKey.id } });
   }
 });
--- a/src/__tests__/security/articles-update.test.ts
+++ b/src/__tests__/security/articles-update.test.ts
@@ -1,0 +1,80 @@
+import assert from "node:assert/strict";
+import { readFile } from "node:fs/promises";
+import path from "node:path";
+import test from "node:test";
+
+import { NextRequest } from "next/server";
+import { _redisTestHooks } from "@/lib/cache/redis";
+import { FakeRedisClient } from "../_helpers/fake-redis";
+
+test.beforeEach(() => {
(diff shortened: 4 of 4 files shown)


SweetSophia (commented):
# Review — PR #237: test: cover article update rate and body limits

**Verdict: Request changes** (1 must-fix, 1 should-fix) — not because the tests are buggy, but because the PR's headline security assertion won't actually run in CI, and one test is redundant/brittle.

Well-written overall. The production route's security properties (rate-limit-before-auth + bounded JSON parser) genuinely hold and are correctly characterized by the tests. The substantive items are about *enforcement scope* and *test maintainability*, not correctness.

## ✅ Verified correct
- `PATCH /api/articles/[id]` (`src/app/api/articles/[id]/route.ts`): `rateLimit({maxRequests:30, keyPrefix:"articles-patch"})` runs **before** `requirePermission` (L39 vs L44), and the body is parsed via `readBoundedJsonObject(request, ARTICLE_JSON_BODY_MAX_BYTES)` (L81). The tests describe real behavior.
- **Test A is deterministic, not flaky:** fresh `FakeRedisClient` per test, unique `now:UUID` ZSET members, runtime ≪ the 60s window. The 30-allowed-then-429 pattern is sound, and **the 31st → 429 is what actually proves limiter-before-auth ordering** (if auth ran first and short-circuited to 401, the limiter counter would never advance and you'd never observe 429).
- **The 413 integration test is a precise oracle.** The body (~1.13 MB = `maxContentSize` 1 MB + `DEFAULT_JSON_BODY_MAX_BYTES` 64 KB + 1) exceeds `ARTICLE_JSON_BODY_MAX_BYTES`, so `RequestBodyTooLargeError → 413`. If the route ever regressed to `request.json()`, App Router imposes no default size cap on it, so this body would be read fine and then rejected by the route's own `content > maxContentSize` guard as **400, not 413**. So 413 definitively proves the bounded parser is the active path.
- `.env.test` only sets `DATABASE_URL`, so broadening `--env-file=.env.test` to the existing `uploads`/`proxy` security tests is harmless.

## 🔴 Must-fix — CI doesn't run `test:security`
`test:security` is excluded from `.github/workflows/ci.yml` — the test job runs only `test:persistence-layer` and `test:api`. So **Test A, the PR's central "rate-limits before auth" assertion, is enforced only when a human remembers to run `npm run test:security` locally.** The 413 test does run in CI (it lives under `src/__tests__/api/`). CI is the merge gate; a security-only test that doesn't run there is a review artifact, not enforcement.

Inclusion is free and safe:
- **No Redis/infra needed** — `beforeEach` injects a `FakeRedisClient` whose `connect()` flips `status="ready"`.
- **No DB queries** — Test A's unauthenticated requests short-circuit at `requireApiKey` (no `Authorization` header), then `getServerSession` throws in the bare harness → 401. No SQL fires in any of the 31 iterations.
- **Prisma client construction is lazy** — only needs `DATABASE_URL` to be present, which the CI postgres job already sets.

**Recommendation:** add `npm run test:security` to the `test:` job in `ci.yml`, after `test:api`.

*Defense-in-depth note:* the middleware (`src/proxy.ts`) already rate-limits this same route via the `api-write` bucket (covered by `src/__tests__/security/proxy.test.ts`), and Test A locks the **route-level** `articles-patch` bucket independently. Both layers are worth keeping — but only if the route-level test actually runs in CI.

## 🟡 Should-fix — Test B is redundant and brittle
`"PATCH /api/articles/[id] uses the bounded JSON parser"` greps the route's source text:

```ts
assert.doesNotMatch(source, /request\.json\s*\(/);
assert.match(source, /readBoundedJsonObject<[^>]+>\s*\(\s*request,\s*ARTICLE_JSON_BODY_MAX_BYTES,\s*\)/);
```

- **Redundant** — the 413 integration test above already proves the bounded parser is used *and enforced*, more reliably.
- **Brittle to cosmetic refactors** — the positive regex couples to the literal identifier `ARTICLE_JSON_BODY_MAX_BYTES` and the generic `<...>` form. Trivial, invariant-preserving changes break it: extracting a local (`const cap = ARTICLE_JSON_BODY_MAX_BYTES`), inserting an inline comment between args (`request /* x */, ...`), or dropping the explicit type parameter.
- **Negative regex matches comments/strings too** — a future `// avoids request.json()` explanatory comment would fail the test.
- **cwd-dependent path** — `path.join(process.cwd(), "src/app/api/articles/[id]/route.ts")` only works when tests are invoked from the repo root. If kept, resolve via `import.meta.dirname`.

**Recommendation:** delete Test B. If a source-level guard is desired, scope the negative match to code (not comments) and resolve the path relative to the test file.

## 🟡 Nice-to-have
- **Duplicate magic `30`.** The route's `maxRequests: 30` (`route.ts:39`) and Test A's loop bound (`i < 30`) aren't linked. Tuning the route silently breaks the test. Consider exporting a small `ARTICLES_PATCH_RATE_LIMIT` constant and referencing it from the test.
- **Comment precision (Test A).** *"this assertion never reaches DB work"* is true in effect but misleading in mechanism — the first 30 *do* enter `requirePermission`; they avoid SQL only because `requireApiKey` short-circuits (no Bearer) and `getServerSession` throws in the bare harness. Suggested rewrite: *"The first 30 pass the limiter and reach `requirePermission`, which returns 401 with no DB work. The 31st returning 429 is what proves the limiter runs before auth."*
- **PR description vs. code (nit).** The description says the integration reorder "ensures audit log assertions happen before deletion," but the diff moves `assert.deepEqual(auditLogs, …)` to *after* the deletes. Functionally equivalent (`auditLogs` is a captured array), but the wording is inaccurate — either revert the reorder or fix the description.

## Out of scope (not introduced here)
The rate limiter **fails open** when Redis is unavailable (`src/lib/rate-limit.ts:31-33, 59-62`) — an attacker who can disrupt Redis bypasses both the proxy and route-level limiters, leaving `requirePermission`/`validateApiKey` (an indexed DB lookup per call) as the only brake. This is a documented system-wide design tradeoff, not something this PR changes. Worth a separate issue if not already tracked.

---
*Review reconciled from an independent security/correctness pass and first-principles source verification. Happy to send a follow-up that wires `test:security` into CI.*

SweetSophia:
Addressed the review feedback in `c5ed0b2`:

- Added `npm run test:security` to `.github/workflows/ci.yml`, after the API test step, so the article PATCH route-level rate-limit test now runs in CI.
- Removed the redundant/brittle source-grep bounded-parser assertion from `articles-update.test.ts`; bounded-parser coverage remains enforced behaviorally by the existing oversized PATCH integration test returning 413.
- Tightened the remaining rate-limit test comment to describe the 30x401 + 1x429 proof accurately.
- Refreshed the PR description so it no longer mentions the removed source-text test.

Local verification:
- `npm run test:security` — 21/21 passing
- `npm run test:api` — 350/350 passing against an isolated temporary Postgres 16 database
- `npm run lint` — pass, with the existing unrelated `descendants` warning
- `npm run typecheck` — pass
- Two sequential follow-up review passes — no blockers

SweetSophia (commented):
# Final re-review — PR #237

**Verdict: ✅ Ready to merge.** Both blocking items from the first review are resolved and verified green in CI on `c5ed0b2`.

## Resolved
- 🔴 **`test:security` now runs in CI** — `.github/workflows/ci.yml` adds a `Run security tests` step (`npm run test:security`) after `Run API tests`, in the postgres-backed job. The `test` check is **SUCCESS**, so the route-level "rate-limits before auth" assertion is now actually enforced at the merge gate. No Redis/infra needed (fake client) and no live queries (unauth short-circuits before any SQL) — inclusion is free, as expected.
- 🟡 **Test B removed** — `articles-update.test.ts` is now Test A only. No dangling imports (`lint-and-typecheck` SUCCESS confirms). The 413 integration test remains the authoritative oracle for the bounded parser.
- 🟡 **Comment precision fixed** — Test A now reads *"The first 30 requests pass the limiter and return 401 without DB work… The 31st returning 429 proves the route-local limiter runs before auth."* Accurate.

## Remaining (optional, non-blocking)
- Magic `30` still duplicated between `route.ts:39` (`maxRequests: 30`) and the test loop (`i < 30`) — fine to defer.
- The audit-log reorder in the integration test persists; functionally equivalent (`auditLogs` is a captured array), purely cosmetic.

## Verification
Diff-only delta (3-line CI addition + test deletion + comment edit) confirmed by:
- `test` check SUCCESS — `test:security` passes in CI (incl. the previously-local `proxy`/`uploads` security tests, which now also gain CI coverage as a side benefit).
- `lint-and-typecheck` SUCCESS — `crypto.randomUUID()` resolves via the Node 22 global + `@types/node`, no missing import; no unused imports after Test B's removal.

No new issues introduced. Ship it. 🚢

PR #238, opened 2026-07-04
Fix bootstrap generated secret logging

SweetSophia:
## What changed

- **`docker/bootstrap.mjs`** — When the bootstrap init container generates credentials (admin password and/or API key) because the operator did not supply them, it now:
  - Persists them to a JSON file at `NOOSPHERE_BOOTSTRAP_SECRETS_FILE` (default `/tmp/noosphere-bootstrap-secrets.json`) inside the container.
  - Writes the file atomically via a temp file + rename, and enforces `0600` permissions on both the temp file and the final path.
  - Logs only the file path on stderr — never the raw credential values.
  - Moves the `COMMIT` after secret handling so the JSON summary still reports auto-generation status plus the secrets file path.
- **`docker-compose.noosphere.yml`** — Forwards `NOOSPHERE_BOOTSTRAP_SECRETS_FILE` (defaulting to the container-local `/tmp/noosphere-bootstrap-secrets.json`) to the init service.
- **`install-openclaw.sh`** — Threads `NOOSPHERE_BOOTSTRAP_SECRETS_FILE` through the generated `.env`, the exported variables consumed by `docker compose`, and the inline compose template emitted by the installer.
- **`noosphere.env.example`** — Documents the new variable with a note about pointing it at the `uploads` volume (`/app/uploads/bootstrap-secrets.json`) for persistence past init-container teardown.
- **`README.md`** — Adds a quickstart note explaining the default location, the `0600` mode, the "path only, not the secrets" logging behavior, and how to override the path.
- **`docs/OPENCLAW-OFFICIAL-PLUGIN-SETUP.md`** — Adds `NOOSPHERE_BOOTSTRAP_SECRETS_FILE` to the installer environment-variable table.
- **`src/__tests__/api/bootstrap-topics.test.ts`** — Adds a regression test asserting that `docker/bootstrap.mjs` references `NOOSPHERE_BOOTSTRAP_SECRETS_FILE` and `writeGeneratedSecretsFile`, that it no longer interpolates `adminPassword`/`rawApiKey` into log strings, and that both the compose template and the installer wire the variable through.

## Why

Previously, when bootstrap auto-generated the admin password or API key, the raw values were written to stdout/stderr via `console.error` and then surfaced through `docker compose up` logs. Anyone with log access could recover full administrative credentials. The fix routes auto-generated secrets into a `0600` file (and out of logs) so they remain accessible to the operator without leaking into shared log streams, while preserving the existing behavior when operators supply credentials via environment variables.

---

## Description

This pull request fixes a security issue in the bootstrap init container where auto-generated admin passwords and API keys were being written directly to the logs (stderr). Secret material should never appear in container logs.

### Changes

- **Removed secret values from log output**: The bootstrap script no longer prints the generated admin password or API key to stderr. Instead, when credentials are auto-generated, it writes them to a JSON file at `/tmp/noosphere-bootstrap-secrets.json` (inside the init container) with restrictive `0600` file permissions and `0700` directory permissions, and only logs the file path.
- **Atomic file writes with preserved existing credentials**: Secrets are written via a temp file + rename to avoid partial writes, and any credentials already present in an existing secrets file are merged so that repeated bootstrap runs (e.g. when `depends_on` re-runs the init service) don't drop previously generated values.
- **Configurable file path**: A new `NOOSPHERE_BOOTSTRAP_SECRETS_FILE` environment variable lets operators point the secrets file to a different container path (for example, a mounted volume like `/app/uploads/bootstrap-secrets.json`) to persist generated credentials past container teardown.
- **Transaction safety fix**: The script now tracks whether a DB transaction is open before attempting rollback in the error path, avoiding spurious errors after a successful commit.
- **Module export guard**: `main()` is now only invoked when the script is run directly, making the helpers testable as an ES module.
- **Installer / compose / docs / example env updates**: The new `NOOSPHERE_BOOTSTRAP_SECRETS_FILE` variable is wired through `docker-compose.noosphere.yml`, the OpenClaw installer (`install-openclaw.sh`), `noosphere.env.example`, the README, and the OpenClaw setup docs so the behavior is consistent across deployment paths.
- **Tests**: Added unit tests covering the new `collectGeneratedSecrets` and `writeGeneratedSecretsFile` helpers (file mode, merging behavior, skipping when nothing is generated) plus a static test asserting the bootstrap, compose, and installer sources no longer embed raw secret values in logs.

---

## Summary

This pull request hardens the handling of bootstrap-generated secrets and improves the safety of bootstrap error output.

### What's changed

- **Secrets file path hardened to require a dedicated private directory.** The bootstrap script now refuses to write `NOOSPHERE_BOOTSTRAP_SECRETS_FILE` to a path whose parent is a shared directory such as `/tmp` or `/app/uploads`. Only two roots are accepted: `/tmp/noosphere-bootstrap-secrets` and `/app/uploads/bootstrap-secrets`. The default path is updated to `/tmp/noosphere-bootstrap-secrets/secrets.json`, and the parent directory is always created (or re-tightened) with mode `0700` before the `0600` credentials file is written inside it.

- **Bootstrap error output sanitized.** When bootstrap fails, the script no longer prints the raw exception message. Instead it emits a fixed-format line that includes only the error name and (when present) an error code, with a hint to verify database and bootstrap configuration. This prevents accidental leakage of supplied secrets, database URLs, or other sensitive values via stack traces or driver error messages.

- **Documentation and configuration aligned with the new contract.** `README.md`, `docs/OPENCLAW-OFFICIAL-PLUGIN-SETUP.md`, `noosphere.env.example`, `docker-compose.noosphere.yml`, and the generated `docker-compose.yml` in `install-openclaw.sh` all point at the new default path and explain that operators should set `NOOSPHERE_BOOTSTRAP_SECRETS_FILE=/app/uploads/bootstrap-secrets/secrets.json` to persist the file inside the `noosphere_uploads` volume.

- **Tests added.** The bootstrap test suite now verifies that the parent directory ends up at mode `0700`, that previously loose pre-existing files are re-tightened to `0600` on the next write, that the helper rejects shared parent directories (including paths that traverse out via `..`), and that running the bootstrap script against an unreachable database produces output containing none of the supplied admin password, API key, or database password while still reporting a structured failure line.

---

This pull request fixes a potential secret-leakage issue in the bootstrap script's error handling.

**Problem**: When `docker/bootstrap.mjs` encountered a missing required environment variable (such as `DATABASE_URL`) or an invalid `NOOSPHERE_BOOTSTRAP_SECRETS_FILE` path, the generic `Error` object could carry arbitrary message content. The previous error formatter unconditionally included the error message in its output, which risked exposing sensitive context (for example, parts of file paths or other details) alongside bootstrap output.

**Changes**:
- Introduces a `SafeBootstrapError` class to mark errors whose messages are safe to surface in error logs (currently: missing required env vars and invalid secrets-file parent-directory validation).
- The error formatter now only appends the original `error.message` when the error is an instance of `SafeBootstrapError`. Any other error type still produces a sanitized message that contains no error detail at all, preventing unintended information leakage.
- Adds a regression test (`bootstrap process output keeps safe validation context without supplied secrets`) that:
  - Spawns the bootstrap script with intentionally missing `DATABASE_URL` while supplying a real-looking admin password and API key.
  - Asserts that the missing-env-var message (`DATABASE_URL is required`) is surfaced.
  - Asserts that the supplied admin password and API key do not appear anywhere in stdout/stderr.

**Functional impact**: Operators now get actionable diagnostic context for the two safe-to-expose failure modes (missing env vars, invalid secrets-file location), while still being guaranteed that secret values are never echoed to bootstrap logs for any other failure path.

---

Use a non-loggable error type for invalid permission validation in bootstrap script, and add a test verifying that the safe permission error message is shown without leaking secrets.


SweetSophia:
Addressed the review feedback in commit `646e438`:

- The final secrets file is now written only after `COMMIT` succeeds, so failed transactions do not leave authoritative-looking credentials behind.
- Removed the redundant final `chmodSync` after `renameSync`; the temp file is created with `0600` and renamed with that mode preserved.
- Installer reruns now recover an existing `NOOSPHERE_BOOTSTRAP_SECRETS_FILE` from the runtime `.env`, preserving custom paths.
- Added behavioral tests for generated-secret helper behavior: mode `0600`, JSON content, missing-directory creation, empty/unapplied credential omission, and preservation of existing credentials during reruns.

Re-verified locally:
- `node --import tsx --test src/__tests__/api/bootstrap-topics.test.ts` — 5/5 passing
- `npx eslint docker/bootstrap.mjs src/__tests__/api/bootstrap-topics.test.ts` — passing
- `bash -n install-openclaw.sh` — passing
- `npm run typecheck` — passing
- `npm run lint` — passing with the pre-existing unrelated `src/app/wiki/admin/topics/actions.ts` warning
- `git diff --check` — clean
- Compose config confirms `NOOSPHERE_BOOTSTRAP_SECRETS_FILE` reaches init
- Behavioral bootstrap smoke against temporary Postgres: `0600` file, generated credentials omitted from stdout/stderr

SweetSophia:
## In-depth review — PR #238 (Fix bootstrap generated secret logging)

**Verdict: approve with suggestions.** The core fix is correct and meaningfully reduces the secret-exposure surface — auto-generated `adminPassword` and `rawApiKey` no longer enter stderr, they land in a `0600` JSON file with atomic write+rename, and the JSON summary still surfaces which credentials were auto-generated. No **must-fix** items. Two items **strongly recommended before merge** (defense-in-depth); the rest are nice-to-have.

### What this PR gets right

- **Least-privilege channel split.** Container logs (docker logs, journald, log drivers) are an aggregation surface where credentials have no business living; routing them to a `0600` file on a mounted volume is the right move. (anchor: `least-privilege`)
- **Atomic write + rename with `0o600`.** A crashed mid-write init can't leave a half-written plaintext file; the destination inherits the temp file's already-tightened inode perms. (anchor: `atomicity`)
- **Path-only stderr logging.** Even the existence/rotation banner only logs the file path. This is the single most important detail and it's done right. (anchor: `principle-of-least-disclosure`)
- **`transactionOpen` flag** is the correct way to fix the rollback bookkeeping — a boolean gate beats inspecting pool state.
- **ESM testability split** — `if (import.meta.url === file://${process.argv[1]})` is the canonical Node idiom and lets the suite import `collectGeneratedSecrets`/`writeGeneratedSecretsFile` without auto-invoking `main()`.
- **Fail-secure by default:** missing credentials now fall back to a non-log file rather than `console.error`, which removes the obvious path to credential exfiltration.
- **Belt-and-suspenders tests** — both a static grep guard *and* a behavioral round-trip; together they cover most regression shapes.

### Strongly recommended before merge

**SR1 — Parent dir mode not tightened on pre-existing volume.** `docker/bootstrap.mjs`, `writeGeneratedSecretsFile`: `mkdirSync(parentDir, { recursive: true, mode: 0o700 })` only applies the mode to directories Node *creates*. On a pre-existing `/app/uploads` named volume (mode inherited from whoever mounted it first; often `0755`), the secrets file's parent dir stays world-*traversable*. The `0o600` file itself remains unreadable to other users, but they can stat it, see it exists, infer the file's purpose, and enumerate the volume. One-line fix: add an unconditional `chmodSync(parentDir, 0o700)` after the `mkdirSync`. (anchor: `fail-secure`, `defense-in-depth`)

**SR2 — Test gate is bypassable by trivial refactor.** `src/__tests__/api/bootstrap-topics.test.ts`, first new test. The static `doesNotMatch(/Auto-generated admin password: \$\{adminPassword\}/)` only catches exactly the templated literal that exists today. A motivated refactor can defeat it without restoring the leak by: splitting the log across two `console.error` calls, routing through a helper, switching to `process.stdout.write`, or renaming the variable. The behavioral test that already exists for `writeGeneratedSecretsFile` proves the mechanics; it does not prove the *regression* is caught end-to-end. Replace/augment with a behavioral test that spawns `node docker/bootstrap.mjs` in a tempdir, captures stderr+stdout, and asserts the password/key strings never appear in either stream. Keep the grep as belt-and-suspenders. (anchor: `brittleness-of-static-assertions`, `completeness-of-mediation`)

### Suggested before merge

**S1 — `error.message` may carry the connection string.** `docker/bootstrap.mjs`, top-level catch. The current `console.error(error instanceof Error ? error.message : String(error))` is a known anti-pattern: `pg`, `bcrypt`, and a host of other libraries can include config / DSN data (and therefore `POSTGRES_PASSWORD`) in their error messages. The bootstrap fix already moves *credentials* out of stderr; this moves the *credential-bearing context* back in. Fix: log `{ name: error.name, code: error.code }` plus a fixed diagnostic string, exit non-zero. Operators on the host have the secrets file for status. This is out of scope for "fix the log leak" but the surrounding diff touches the line, so worth including. (anchor: `fail-secure`)

**S2 — README quickstart paragraph is missing the persistence caveat.** `README.md`. The new paragraph covers default location, `0600` mode, path-only, and override, but does not explicitly say *the default path lives inside the init container's filesystem and is destroyed on container exit*. A first-time reader running `docker compose up -d` and then `docker compose logs init` won't connect "default `/tmp/...`" → "lost on next deploy." One-sentence fix: *"Default `/tmp/...` is inside the init container and is destroyed when it exits; point the var at `/app/uploads/bootstrap-secrets.json` to persist across restarts (the `noosphere_uploads` volume is mounted read-write on both `init` and `app`)."* `noosphere.env.example` already hints at this and is OK.

### Nice-to-have follow-ups

- **S3 — TOCTOU between `mkdirSync` and `writeFileSync`** (defense-in-depth only): if `secretsFile` points into a writable shared dir, an attacker who wins the race pre-places a symlink at the tmp-name. Inside an init container this is largely hypothetical (single-writer, ephemeral), but `mkdtempSync` gives atomicity + symlink safety in one move. Replace the hand-rolled `${pid}.${rand}.tmp` suffix with `path.join(mkdtempSync(...), "secrets.json")` then rename into place.
- **S4 — `createdAt` is overwritten on every merge-write.** The "first-generation" timestamp is lost on any subsequent merge. Fix: `createdAt: existing?.createdAt ?? new Date().toISOString()`, optionally add `lastUpdatedAt`.
- **S5 — Redundant `chmodSync(0o600)` after `writeFileSync({mode: 0o600})`.** Cheap, keep, but a one-line comment would prevent a future reader deleting it as dead code.
- **S6 — Installer `.env` template lacks a default** for the new var. The `env_get_secret "$NOOSPHERE_HOME/.env" NOOSPHERE_BOOTSTRAP_SECRETS_FILE` line returns empty on older installs' rerun. The downstream compose uses YAML `${VAR:-/tmp/...}` so this is harmless in the `init` path, but a `:?` guard or default in the installer's `.env` template would make the export-list wiring self-consistent.
- **S7 — Heredoc escaping in `install-openclaw.sh`** (~lines 426-428). The `\${NOOSPHERE_BOOTSTRAP_SECRETS_FILE:-/tmp/noosphere-bootstrap-secrets.json}` only expands correctly because the heredoc EOF marker is unquoted. A future refactor that quotes the marker breaks docker-compose. One-line comment in the script would prevent this footgun.
- **S8 — Installer success banner doesn't mention the bootstrap secrets file.** `install-openclaw.sh:621-644` shows the OpenClaw `SECRETS_FILE` but not `NOOSPHERE_BOOTSTRAP_SECRETS_FILE`. The operator who runs the installer learns about the OpenClaw file via "Credentials also saved in: …" but has no in-band clue about the bootstrap file's separate existence or the persistence caveat. (Note: in installer mode, the bootstrap secrets file is largely redundant — the installer generates its own `ADMIN_PASSWORD`/`API_KEY` at lines 397-398 and writes them itself to `$SECRETS_FILE`. The bootstrap file matters primarily for the *standalone* `docker compose -f docker-compose.noosphere.yml` flow.) A one-line note in the success banner — *"If you didn't supply `NOOSPHERE_ADMIN_PASSWORD`/`NOOSPHERE_BOOTSTRAP_API_KEY`, generated credentials are in `${NOOSPHERE_BOOTSTRAP_SECRETS_FILE:-/tmp/...}` inside the (now-stopped) init container"* — closes the loop for the standalone case.

### Test gaps worth filling (non-blocking)

- **Re-tighten on existing file:** pre-create `secrets.json` at `0o644`, call writer, assert resulting mode is `0o600` (exercises the rename-source-keeps-mode property).
- **`transactionOpen` regression test:** monkey-patch `users.upsert` to throw after topic COMMIT; assert no `ROLLBACK` is attempted.
- **`createdAt` preservation:** after SR4 lands, assert re-writing preserves the original timestamp.
- **Path-with-spaces / relative path** is cheap and covers a footgun class.
- **Module top-level side-effect audit (T3 from `@oracle` pass):** ✅ the test import via `pathToFileURL` is safe — `Pool` is constructed only inside `main()`, no DB connection at module load. Confirmed separately.

### Cross-surface verification

- `grep -rn 'console\.(log|error|warn)\([^)]*\$\{(adminPassword|rawApiKey|…)\}'` across the repo returns **only the two lines this PR removes** (`docker/bootstrap.mjs:133,136`). The fix is targeted correctly — no other silent credentials-in-logs surfaces remain.
- `install-openclaw.sh:332` (`chmod 700 $SECRETS_DIR`) and `:555` (`install -m 600`) establish precedent for `0o700`/`0o600` on bootstrap-output paths; the PR is consistent with existing conventions.
- `install-openclaw.sh:529` captures bootstrap combined stdout/stderr to `BOOTSTRAP_TMP` and dumps it to the installer's stderr on failure (`:544`). Once this PR lands, that failure-dump is also clean — bootstrap no longer emits raw credentials to either stream, so `cat "$BOOTSTRAP_TMP" >&2` doesn't leak them. Good unintended consequence of the chain of fixes.
- `docker-compose.noosphere.yml` does not mount `/tmp` to a persistent volume; default `NOOSPHERE_BOOTSTRAP_SECRETS_FILE=/tmp/...` is ephemeral until overridden. `noosphere_uploads:/app/uploads:rw` *is* mounted on both `init` and `app`, which is why pointing the var at `/app/uploads/bootstrap-secrets.json` is the canonical persistence recipe.

### Final

- No must-fix items; the secrets are out of logs, the file is `0600`, and the write is atomic.
- Strongly recommended: SR1 (one-line `chmodSync`), SR2 (swap the brittle grep for a behavioral stderr-capture test).
- Everything else is hardening/consistency polish that can land as a follow-up if you'd rather keep this PR tightly scoped to the security fix.

Solid work — thanks for the surgical scope and the test for the merge-preserve case.

SweetSophia:
Follow-up pushed in f98a264.

Addressed the review items:
- SR1: bootstrap generated secrets now default to /tmp/noosphere-bootstrap-secrets/secrets.json and only allow writes under dedicated private roots (/tmp/noosphere-bootstrap-secrets or /app/uploads/bootstrap-secrets) before chmodding the parent to 0700.
- SR2: added behavioral coverage for rejected shared-parent paths and for process output not leaking supplied admin/API/database sentinels on bootstrap failure.
- S1: bootstrap failure logging now emits a sanitized fixed diagnostic with only error name/code metadata.
- S2: README, env example, compose template, installer, and OpenClaw setup docs now document the ephemeral default path and persistent /app/uploads/bootstrap-secrets/secrets.json override.

Local verification before push:
- node --import tsx --test src/__tests__/api/bootstrap-topics.test.ts: 7/7 pass
- npx eslint docker/bootstrap.mjs src/__tests__/api/bootstrap-topics.test.ts: clean
- bash -n install-openclaw.sh: clean
- npm run typecheck: clean
- npm run lint: 0 errors, 1 pre-existing unrelated warning in src/app/wiki/admin/topics/actions.ts
- git diff --check: clean
- docker compose -f docker-compose.noosphere.yml config confirms the new default secrets path

SweetSophia:
Follow-up to the Kilo warning pushed in bc5bea5.

I kept the strict redaction boundary for generic runtime/database errors, but added a narrow safe-validation path for bootstrap-owned errors:
- missing env vars now surface safe context such as `DATABASE_URL is required`
- invalid bootstrap secrets paths can surface the safe `NOOSPHERE_BOOTSTRAP_SECRETS_FILE` validation message
- pg/runtime errors still do not print raw `error.message`, so connection strings and secret-bearing context stay out of stdout/stderr

Additional local verification after this patch:
- node --import tsx --test src/__tests__/api/bootstrap-topics.test.ts: 8/8 pass
- npx eslint docker/bootstrap.mjs src/__tests__/api/bootstrap-topics.test.ts: clean
- bash -n install-openclaw.sh: clean
- npm run typecheck: clean
- npm run lint: 0 errors, same unrelated `descendants` warning
- git diff --check: clean

SweetSophia:
## Re-review — PR #238 (after f98a264 + bc5bea5)

**Verdict: approve.** All four strongly-recommended/suggested items from the first pass are correctly addressed; two are stronger than the original ask. 8/8 tests pass, lint clean, no regressions in the surrounding flow. One minor new finding (N1, below) and a few deferred nice-to-haves — none blocking.

### Verification of prior findings

| ID | Status | Notes |
|----|--------|-------|
| **SR1** Parent dir not tightened on pre-existing volume | ✅ Fixed (stronger) | Replaced my one-line `chmodSync` suggestion with an **allowlist + dedicated dir** design: `ALLOWED_BOOTSTRAP_SECRETS_PARENT_ROOTS = ["/tmp/noosphere-bootstrap-secrets", "/app/uploads/bootstrap-secrets"]`, `ensurePrivateSecretsParent` resolves + prefix-matches the parent, then `mkdirSync(0o700)` + unconditional `chmodSync(0o700)`. This closes SR1 **and** the S3 TOCTOU-on-shared-dirs concern in one move — an operator can't point the var at `/tmp/foo.json` or `/app/uploads/foo.json` directly, so the symlink-attack surface disappears. The test at `bootstrap-topics.test.ts:258` ("rejects shared parent directories") covers three rejection cases including a `/tmp/../tmp/...` lexical-traversal attempt. Good defense-in-depth. (anchors: `fail-secure`, `defense-in-depth`, `completeness-of-mediation`) |
| **SR2** Static grep test bypassable | ✅ Fixed | Two new behavioral tests spawn `node docker/bootstrap.mjs` with sentinel values for `adminPassword`, `apiKey`, and `databasePassword`, then assert none of the sentinels appear in captured stdout+stderr. Test 1 (`bootstrap-topics.test.ts:283`) hits a real ECONNREFUSED via `127.0.0.1:1` and asserts the safe formatter output `(name=Error, code=ECONNREFUSED)`. Test 2 (`:318`) sets `DATABASE_URL=""` and asserts the safe-context carve-out surfaces `DATABASE_URL is required` while still suppressing the admin/api sentinels. This is exactly the behavioral stderr-capture I asked for; the static grep test stays as belt-and-suspenders. (anchor: `brittleness-of-static-assertions` — resolved) |
| **S1** `error.message` may carry `DATABASE_URL` | ✅ Fixed (with carve-out) | New `formatBootstrapError(error)` always emits the fixed prefix `[bootstrap] Failed to initialize Noosphere`, adds safe `(name=X, code=Y)` metadata, and surfaces `error.message` **only** for `SafeBootstrapError` instances. `requireEnv` and `ensurePrivateSecretsParent` are the only `SafeBootstrapError` throw sites; both messages are bootstrap-owned and secret-free. pg/bcrypt/fs errors have their messages redacted (only name+code surface). The carve-out is well-scoped — bootstrapped-owned validation errors reach the operator, runtime errors don't. |
| **S2** README/docs missing persistence caveat | ✅ Fixed | All 5 docs surfaces (README, `noosphere.env.example`, `docs/OPENCLAW-OFFICIAL-PLUGIN-SETUP.md`, `docker-compose.noosphere.yml`, `install-openclaw.sh` inline template) now consistently document: default path is `/tmp/noosphere-bootstrap-secrets/secrets.json`, mode `0600` inside `0700` parent, destroyed on init-container exit, override with `/app/uploads/bootstrap-secrets/secrets.json` for persistence, paths under shared dirs are rejected. The new comment block above the inline compose template line explaining *why* the default lives in a subdirectory is exactly the kind of context that prevents a future refactor from breaking the constraint. |

### New finding (minor, non-blocking)

**N1 — Permission-validation error not promoted to `SafeBootstrapError`.** `docker/bootstrap.mjs:161-163`:
```js
if (!["READ", "WRITE", "ADMIN"].includes(permissions)) {
  throw new Error("NOOSPHERE_API_KEY_PERMISSIONS must be READ, WRITE, or ADMIN");
}
```
This is a pre-existing throw that the S1 carve-out didn't reach. The message is safe (env-var name + enum, no secret material), but because it's a plain `Error` it now gets redacted by `formatBootstrapError`. An operator who sets `NOOSPHERE_API_KEY_PERMISSIONS=SUPERUSER` sees only `[bootstrap] Failed to initialize Noosphere (name=Error). Verify database and bootstrap configuration.` with no hint of the actual cause, even though surfacing it would be safe. Fix: `throw new SafeBootstrapError("NOOSPHERE_API_KEY_PERMISSIONS must be READ, WRITE, or ADMIN")`. One-line change, restores the diagnostic without weakening the redaction boundary. (This is the only `throw new Error(...)` in the file that contains a safe, operator-actionable message.)

### Deferred nice-to-haves (still open, low priority)

These were flagged as nice-to-haves in the first pass and remain unaddressed; none block merge:

- **S4 — `createdAt` overwritten on every merge-write.** `docker/bootstrap.mjs:99` still uses `createdAt: new Date().toISOString()` unconditionally. A re-run that merges into an existing file loses the original generation timestamp. Low audit value but trivial to fix: `createdAt: existing?.createdAt ?? new Date().toISOString()`.
- **S5 — Redundant `chmodSync(tmpFile, 0o600)` after `writeFileSync({mode: 0o600})`.** Cheap defense, keep it, but a one-line comment would prevent a future reader deleting it as dead code.
- **S6 — Installer `.env` template lacks an explicit default** for `NOOSPHERE_BOOTSTRAP_SECRETS_FILE`. The compose YAML's `${VAR:-default}` fallback covers the runtime path, so this is functionally harmless, but the wiring would be cleaner with a default in the template.
- **S8 — Installer success banner doesn't mention the bootstrap secrets file.** `install-openclaw.sh:621-644` still only references the OpenClaw `$SECRETS_FILE`. In installer mode this is largely redundant (the installer writes its own secrets file at `:555-565`), but operators using the standalone `docker compose -f docker-compose.noosphere.yml` flow have no in-band clue about `NOOSPHERE_BOOTSTRAP_SECRETS_FILE`. A one-line note in the success banner would close the loop for that case.

### Verification

- ✅ `node --import tsx --test src/__tests__/api/bootstrap-topics.test.ts` → 8/8 pass (confirmed locally)
- ✅ `npx eslint docker/bootstrap.mjs src/__tests__/api/bootstrap-topics.test.ts` → clean
- ✅ The two behavioral spawn-tests use sentinel values (`sentinel-admin-password-output-test`, etc.) and assert non-presence in captured output — the regression guard is real, not theatrical
- ✅ The allowlist approach correctly rejects `/tmp/noosphere-bootstrap-secrets-rejected.json`, `/tmp/../tmp/...`, and `/app/uploads/bootstrap-secrets-rejected.json` (verified by test 6)
- ✅ Path validation runs before `mkdirSync` — no side effects on rejection
- ✅ `chmodSync` runs unconditionally after `mkdirSync` — closes the pre-existing-volume mode gap
- ✅ `SafeBootstrapError` correctly carries safe messages through the redaction boundary; pg/runtime errors stay redacted

### Final

Solid follow-up. The allowlist design is a genuine improvement over my original SR1 suggestion — it closes a class of path-confusion vulnerabilities rather than just patching the one mode bit. The behavioral tests are the right shape and use sentinel values that won't false-positive. Ship it.

N1 is worth a one-line follow-up commit (`throw new SafeBootstrapError(...)` at line 162) but doesn't need to block the merge.

SweetSophia:
N1 follow-up pushed in c6f45f6.

Change:
- `NOOSPHERE_API_KEY_PERMISSIONS` validation now throws `SafeBootstrapError`, so the safe enum diagnostic survives the redaction boundary.
- Added a subprocess regression test for `NOOSPHERE_API_KEY_PERMISSIONS=SUPERUSER` that asserts the sanitized bootstrap prefix + safe validation message are present, while supplied admin/API/database sentinels remain absent from stdout/stderr.

Local verification after the patch:
- `node --import tsx --test src/__tests__/api/bootstrap-topics.test.ts`: 9/9 pass
- `npx eslint docker/bootstrap.mjs src/__tests__/api/bootstrap-topics.test.ts`: clean
- `bash -n install-openclaw.sh`: clean
- `npm run typecheck`: clean
- `npm run lint`: 0 errors, same unrelated `descendants` warning
- `git diff --check`: clean

Two sequential review passes found no must-fix issues.

SweetSophia:
## Re-review — PR #238 (after c6f45f6)

**Verdict: ship it.** N1 is fixed correctly with the right behavioral coverage. 9/9 tests pass, lint clean. No further changes needed from me.

### N1 verification

| Aspect | Status |
|--------|--------|
| `docker/bootstrap.mjs:162` — `throw new Error(...)` → `throw new SafeBootstrapError(...)` | ✅ Exact one-line fix |
| New behavioral test (`bootstrap-topics.test.ts:259`, "keeps safe permission validation context") | ✅ Spawns bootstrap with `NOOSPHERE_API_KEY_PERMISSIONS=SUPERUSER`, asserts the safe enum diagnostic surfaces while admin/api/database sentinels stay redacted |
| `formatBootstrapError` carve-out reaches the permission validation path | ✅ Verified — `SafeBootstrapError` instance triggers the `: ${error.message}` branch |

The test design mirrors the two prior behavioral tests (sentinel values + spawn + assert non-presence in stdout/stderr), so the regression guard is consistent with the existing pattern. An operator who sets an invalid permission now sees `[bootstrap] Failed to initialize Noosphere (name=SafeBootstrapError): NOOSPHERE_API_KEY_PERMISSIONS must be READ, WRITE, or ADMIN. Verify database and bootstrap configuration.` — safe, actionable, no secret leakage.

### Cumulative finding status across all three passes

| ID | Severity | Status |
|----|----------|--------|
| SR1 parent dir mode | strongly recommended | ✅ Fixed (allowlist design — stronger than asked) |
| SR2 brittle grep test | strongly recommended | ✅ Fixed (behavioral spawn tests) |
| S1 `error.message` leak | suggested | ✅ Fixed (SafeBootstrapError carve-out) |
| S2 README persistence caveat | suggested | ✅ Fixed (all 5 docs surfaces) |
| S3 TOCTOU on shared dirs | nice-to-have | ✅ Folded into SR1's allowlist design |
| N1 permission validation diagnostic | minor | ✅ Fixed in c6f45f6 |
| S4 `createdAt` overwritten on merge | nice-to-have | Open — low audit value, trivial follow-up |
| S5 redundant `chmodSync` comment | nit | Open |
| S6 installer `.env` template default | nit | Open — functionally covered by compose `${VAR:-default}` |
| S8 installer success banner mention | nice-to-have | Open — UX gap for standalone compose flow only |

### Verification

- ✅ `node --import tsx --test src/__tests__/api/bootstrap-topics.test.ts` → 9/9 pass (confirmed locally on `c6f45f6`)
- ✅ `npx eslint docker/bootstrap.mjs src/__tests__/api/bootstrap-topics.test.ts` → clean
- ✅ The three behavioral spawn-tests now cover: pg ECONNREFUSED (runtime error redaction), missing DATABASE_URL (env validation carve-out), invalid permission (enum validation carve-out) — the carve-out boundary is tested from both sides
- ✅ All `throw new Error(...)` sites in `docker/bootstrap.mjs` have been audited: the only remaining one is inside `formatBootstrapError` itself (which constructs the safe string), not a throw site

Three review passes, zero must-fix items remaining. The PR meaningfully reduces the secret-exposure surface (auto-generated credentials no longer enter container logs under any code path), the regression guards are behavioral rather than brittle, and the operator UX for bootstrap-owned validation errors is preserved. Approve.

Issue #241, opened 2026-07-08
B4: Verify SSR markdown rendering uses rehype-sanitize + add snapshot test


SweetSophia:
## Source
REVIEW.md §B4

## Problem
The server-rendered markdown path in `src/app/wiki/[topicSlug]/[articleSlug]/page.tsx` was not fully traced during review. Must confirm it also uses `rehype-sanitize` (the client-side `MarkdownPreviewTabs` does).

## Suggested fix
- Verify the SSR render path uses the same sanitized pipeline
- Add a snapshot test that asserts `<script>` and `<iframe>` in markdown renders as text, not HTML

PR #274, opened 2026-07-08
fix: B4 SSR markdown sanitize tests + F46 light-mode contrast fix

SweetSophia:
## Summary

Addresses B4 and F46 with focused changes:

- Adds a shared `WikiMarkdown` renderer that owns the wiki article markdown pipeline: `ReactMarkdown`, `remark-gfm`, `rehype-sanitize`, the article sanitize schema, and the syntax-highlighting `code` mapping.
- Uses the shared renderer in both the SSR article page and the editor preview, so production rendering no longer has duplicated sanitize/plugin wiring.
- Adds B4 security tests that render the same shared component through `renderToStaticMarkup` and verify dangerous HTML/protocols are removed while safe markdown output is preserved.
- Raises light-mode `--color-text-faint` from `#81879b` to `#5a637c`, improving contrast against `#f6f2ea` from 3.20:1 to 5.36:1. Dark mode remains unchanged.

## B4 Coverage

The security suite covers:

- `<script>` tag and inner payload removal
- `<iframe>` removal
- raw HTML link/event-handler non-rendering under the production pipeline
- safe markdown link `href` preservation
- mixed-case `javascript:` URL stripping
- heading, bold, italic, inline code, unordered list, and list item preservation
- syntax-highlighting class preservation with word-boundary assertions
- guard against ReactMarkdown internal `node` props leaking into HTML

## Verification

- `npx eslint src/components/wiki/WikiMarkdown.tsx 'src/app/wiki/[topicSlug]/[articleSlug]/page.tsx' src/components/wiki/MarkdownPreviewTabs.tsx src/__tests__/security/markdown-sanitize.test.ts`
- `npx tsc --noEmit`
- `npm run test:security`
- `DATABASE_URL=<local noosphere db> npm run build`

**Pull Request Summary**

This PR consolidates duplicate markdown rendering logic into a single shared component and hardens SSR sanitization tests, plus fixes a light-mode contrast issue.

**Key Changes**

1. **Extracted shared `WikiMarkdown` component** (`src/components/wiki/WikiMarkdown.tsx`, new): Centralizes the ReactMarkdown + remark-gfm + rehype-sanitize configuration, custom sanitize schema (preserves `className` on `code`/`pre` for syntax highlighting), and the syntax-highlighting `code` component override that was previously duplicated.

2. **Replaced inline markdown setup in two call sites**:
   - SSR article page (`src/app/wiki/[topicSlug]/[articleSlug]/page.tsx`) now uses `<WikiMarkdown content={article.content} />`.
   - Editor preview tabs (`src/components/wiki/MarkdownPreviewTabs.tsx`) now uses `<WikiMarkdown content={trimmedContent} />`.
   
   This ensures both the SSR article rendering and the live editor preview share an identical markdown pipeline, so a future change to sanitization or syntax highlighting can't drift between them.

3. **Hardened B4 SSR sanitization tests** (`src/__tests__/security/markdown-sanitize.test.ts`): The tests now exercise the real `WikiMarkdown` component instead of a duplicate local `ReactMarkdown` instance. Assertions were tightened:
   - `<script>` test now also asserts the script body (`alert("xss")`) is removed, not just the tag.
   - Inline event handler test now also asserts raw `<a>` HTML is not emitted, plus added a positive test that safe markdown links preserve `href`.
   - `javascript:` URL test now uses mixed-case input to catch case-insensitive bypasses and uses a case-insensitive regex.
   - Formatting test now uses a precise regex for `<code>` (avoiding false positives from `<codeblock>`), explicitly checks list elements, and guards against ReactMarkdown internals leaking into output.
   - Code block test now uses word boundaries so the `language-js` assertion can't accidentally match `language-json`.

**Functional Impact**

- The wiki's SSR article page and the editor preview now render markdown through a single, audited pipeline, reducing the risk of sanitization regressions in one path but not the other.
- Sanitization tests now fail when production wiring diverges from the shared component (e.g., if `rehype-sanitize` is removed, the schema is weakened, or ReactMarkdown internals leak into the DOM).
- Mixed-case `javascript:` URLs and raw HTML `<a>` tags with event handlers are now explicitly covered by tests, closing two potential XSS regression vectors.

Note: The PR title also mentions an "F46 light-mode contrast fix," but no corresponding code changes for that item are present in the provided diff.

---

**Title:** fix: B4 SSR markdown sanitize tests + F46 light-mode contrast fix

**Description:**

Expanded server-side markdown sanitize test coverage and fixed SSR rendering of code blocks.

**Changes:**

- **Test exports**: Updated `markdown-sanitize.test.ts` to import `ARTICLE_REHYPE_PLUGINS`, `ARTICLE_REMARK_PLUGINS`, and `ARTICLE_SANITIZE_SCHEMA` directly from `WikiMarkdown.tsx`, enabling direct verification of the shared configuration.

- **New configuration test**: Added a test asserting that the shared markdown pipeline preserves syntax-highlighting className attributes for both `code` and `pre` elements in the sanitize schema.

- **Improved code-block test**: Split the syntax-highlighting test into separate assertions for JavaScript (`language-js`) and JSON (`language-json`) fenced code blocks, ensuring each language class is correctly preserved end-to-end.

- **`WikiMarkdown.tsx` SSR fix**: Refactored the custom `code` component to strip the `node` prop via object spread/delete instead of destructuring it out (avoids passing the non-serializable MDAST `node` through server rendering, which would otherwise emit an extra `node="[object Object]"` attribute into the static markup).

---

## Pull Request Description

This PR updates the B4 SSR markdown sanitize test to strengthen its assertions around how raw HTML containing inline event handlers is handled.

### Changes

**Modified Test: B4 raw HTML with inline event handlers**

The test was retitled and its assertions strengthened to verify that raw HTML containing inline event handlers (e.g., `<a href="..." onclick="alert(1)">link</a>`) is fully stripped from the rendered output. Previously, the test focused only on ensuring that the rendered HTML did not contain the dangerous `onclick` attribute. The updated test now asserts both:

1. The dangerous inline event handler attribute does not appear in the output
2. The raw `<a>` tag itself is also absent, since ReactMarkdown combined with `rehype-sanitize` (without `rehype-raw`) treats raw HTML in the source as text that gets sanitized away

A comment was also added to the test explaining the rendering pipeline behavior so the test's intent is clear to future maintainers.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "test:cache": "node --env-file=.env.test --import tsx --test 'src/__tests__/cache/**/*.test.ts'",
     "test:rate-limit": "node --env-file=.env.test --import tsx --test src/__tests__/rate-limit.test.ts",
     "test:api": "node --import tsx --test 'src/__tests__/api/**/*.test.ts'",
-    "test:security": "node --env-file=.env.test --import tsx --test src/__tests__/security/uploads.test.ts src/__tests__/security/proxy.test.ts src/__tests__/security/articles-update.test.ts",
+    "test:security": "node --env-file=.env.test --import tsx --test src/__tests__/security/uploads.test.ts src/__tests__/security/proxy.test.ts src/__tests__/security/articles-update.test.ts src/__tests__/security/markdown-sanitize.test.ts",
     "lint": "eslint",
     "typecheck": "tsc --noEmit",
     "db:migrate": "prisma migrate dev",
--- a/src/__tests__/security/markdown-sanitize.test.ts
+++ b/src/__tests__/security/markdown-sanitize.test.ts
@@ -3,42 +3,18 @@ import test from "node:test";
 
 import React from "react";
 import { renderToStaticMarkup } from "react-dom/server";
-import ReactMarkdown from "react-markdown";
-import remarkGfm from "remark-gfm";
-import rehypeSanitize, { defaultSchema } from "rehype-sanitize";
+import { WikiMarkdown } from "@/components/wiki/WikiMarkdown";
 
 /**
  * B4: Verify that server-side markdown rendering sanitizes dangerous HTML.
  *
- * The wiki article SSR path (src/app/wiki/[topicSlug]/[articleSlug]/page.tsx)
- * uses ReactMarkdown + remarkGfm + rehype-sanitize with an extended schema
- * that preserves className on code/pre for syntax highlighting.
- *
- * This test asserts that <script>, <iframe>, inline event handlers, and
- * javascript: URLs in markdown content are rendered as inert text,
- * not executable HTML.
+ * These tests render the shared WikiMarkdown component used by the SSR article
+ * page and the editor preview, so they fail if production markdown wiring
+ * drifts away from rehype-sanitize or the shared syntax-highlighting mapping.
  */
 
-const sanitizeSchema = {
-  ...defaultSchema,
-  attributes: {
-    ...defaultSchema.attributes,
-    code: [...(defaultSchema.attributes?.code ?? []), "className"],
-    pre: [...(defaultSchema.attributes?.pre ?? []), "className"],
-  },
-};
-
 function renderMarkdown(content: string): string {
-  return renderToStaticMarkup(
-    React.createElement(
-      ReactMarkdown,
-      {
-        remarkPlugins: [remarkGfm],
-        rehypePlugins: [[rehypeSanitize, sanitizeSchema]],
-      },
-      content,
-    ),
-  );
+  return renderToStaticMarkup(React.createElement(WikiMarkdown, { content }));
 }
 
 test("B4: <script> tags are rendered as inert text, not executable HTML", () => {
@@ -47,6 +23,7 @@ test("B4: <script> tags are rendered as inert text, not executable HTML", () => 
 
   // rehype-sanitize strips the entire <script> element (tag + content)
   assert.doesNotMatch(html, /<script\b/);
+  assert.doesNotMatch(html, /alert\("xss"\)/);
   // Surrounding text is preserved
   assert.ok(html.includes("Some text"), "surrounding text should be preserved");
   assert.ok(html.includes("More text"), "surrounding text should be preserved");
@@ -59,18 +36,27 @@ test("B4: <iframe> tags are rendered as inert text", () => {
   assert.doesNotMatch(html, /<iframe\b/);
 });
 
-test("B4: inline event handlers (onclick) are stripped from links", () => {
+test("B4: raw HTML links with inline event handlers are not rendered as HTML", () => {
   const malicious = '<a href="https://ok.example.com" onclick="alert(1)">link</a>';
   const html = renderMarkdown(malicious);
 
-  assert.ok(!html.includes("onclick"), "onclick must be stripped");
+  assert.doesNotMatch(html, /onclick/i);
+  assert.doesNotMatch(html, /<a\b/);
+});
+
+test("B4: safe markdown links preserve href attributes", () => {
+  const safeLink = "[safe link](https://ok.example.com)";
+  const html = renderMarkdown(safeLink);
+
+  assert.match(html, /href="https:\/\/ok\.example\.com"/);
+  assert.ok(html.includes("safe link"), "safe link text should be preserved");
 });
 
 test("B4: javascript: URLs are stripped from markdown links", () => {
-  const malicious = "[click me](javascript:void(0))";
+  const malicious = "[click me](JaVaScRiPt:void(0))";
   const html = renderMarkdown(malicious);
 
-  assert.ok(!html.includes("javascript:"), "javascript: protocol must be stripped");
+  assert.doesNotMatch(html, /javascript:/i, "javascript: protocol must be stripped");
 });
 
 test("B4: legitimate markdown formatting is preserved", () => {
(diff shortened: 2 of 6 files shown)


SweetSophia:
## PR #274 Review — B4 SSR sanitize tests + F46 contrast fix

Reviewed end-to-end against `fc1e235` on `fix/b4-verify-ssr-sanitize`. CI is green; no merge conflicts.

### TL;DR

Both fixes move in the right direction and address issue **#241 (B4)** correctly. The B4 test coverage is **shallow in the wrong ways**: it exercises a *copy* of the production schema rather than the production wiring, so a regression in `src/app/wiki/[topicSlug]/[articleSlug]/page.tsx` (e.g. someone removing `rehypePlugins` while refactoring) will not trip the tests. This is the dominant architectural concern and was raised independently by Gemini, Kody, Copilot, and Sourcery — I confirm it from the source. The F46 fix is numerically valid but the PR description is factually wrong about which color was used.

### Architecture — Schema duplication (HIGH priority)

`sanitizeSchema` now exists in **three** places:

| File | Lines | Role |
|---|---|---|
| `src/app/wiki/[topicSlug]/[articleSlug]/page.tsx` | 19–26 | SSR article view |
| `src/components/wiki/MarkdownPreviewTabs.tsx` | 15–22 | Client preview/edit |
| `src/__tests__/security/markdown-sanitize.test.ts` | 22–28 | **New** test |

Adding any future attribute allowlist (e.g. `span` for inline color, `details`/`summary` for collapsibles) requires touching all three, and the first two are coupled to React (page is RSC, preview is `"use client"`). Whichever file gets out of sync silently weakens XSS coverage in the forgotten path. `src/lib/markdown/noosphere-markdown.ts` is unrelated (frontmatter parsing) — this needs a new home.

**Recommendation:** extract a shared module, e.g. `src/components/wiki/WikiMarkdown.tsx` (or `src/lib/markdown/wiki.tsx`) exporting:
- `ARTICLE_SANITIZE_SCHEMA`
- `ARTICLE_REMARK_PLUGINS`
- `ARTICLE_REHYPE_PLUGINS`
- `<WikiMarkdown content={article.content} />` (which owns the custom `code` → `SyntaxHighlighter` mapping at `page.tsx:220-241`)

Both `page.tsx` and `MarkdownPreviewTabs.tsx` import the shared module; the new test imports the **same** schema, plugin arrays, and ideally the same `<WikiMarkdown>` rendered through `renderToStaticMarkup`. At that point, the test verifies production wiring instead of a parallel universe.

### B4 test quality — independent findings

The reviewers raised several overlapping concerns; here are the sharpest ones, with my validation:

1. **`onclick` test passes for the wrong reason** (Copilot, sharper than Sourcery). Production rendering does **not** enable `rehype-raw`, so `<a href="..." onclick="...">link</a>` in markdown is never parsed as HTML — `<a>` simply doesn't render, and `onclick` was never an attribute to begin with. The assertion `!html.includes("onclick")` passes because the entire `<a>` tag was discarded. The test as written doesn't verify the security property it claims to. Fix: either enable `rehype-raw` in the test fixture and assert `onclick` is stripped while `href` is preserved, **or** rephrase the test name to reflect what it actually checks.

2. **Custom `components` prop not exercised.** `page.tsx:220-241` overrides `code` to extract `language-(\w+)` from `className` and wrap the block in `<SyntaxHighlighter>`. The test's `renderMarkdown` does not pass this `components` prop, so the `language-js` test verifies that the `language-*` class survives rehype-sanitize, but does **not** verify that the production code component actually picks it up and produces the syntax-highlighted output. Either import the shared component (preferred) or at least add a positive assertion like `assert.match(html, /\blanguage-js\b/)` (note the word boundary — see #3).

3. **`/language-js/` matches `language-json` and `language-jsx`** (Kilo). `assert.match(html, /language-js/)` is a substring search; it lights up green for `language-json`, `language-jss`, `language-jsx`, `language-javascript`, or even an arbitrary `class="contains-language-js-foo"`. Tighten to `assert.match(html, /\blanguage-js\b/)` or `class="[^"]*\blanguage-js\b[^"]*"`.

4. **`javascript:` URL check is case-sensitive** (Kilo, partially valid). `rehype-sanitize` does strip `JavaScript:` case-insensitively (it normalizes protocol casing against its allowlist before comparison), so the assertion will still pass. The HTML-entity variant `&#x6A;avascript:` is also normalized away at the AST level. The case-insensitive variant `assert.doesNotMatch(html, /javascript:/i)` is a defensive improvement, but the gap Kilo describes is largely theoretical given how `rehype-sanitize` works.

5. **`<script>` content re-emission risk** (Sourcery). Defensive addition: also assert the inner text `alert("xss")` is absent. Low probability but cheap.

6. **Test is registered in CI** (chatgpt-codex-connector **incorrect**). The reviewer claimed `test:security` does not enumerate the new file, but `package.json:18` in this PR explicitly adds `src/__tests__/security/markdown-sanitize.test.ts` to the `test:security` script. No action needed; the reviewer's concern is already addressed by the same PR.

### F46 — numerically correct, description wrong

Verified WCAG contrast against the actual light-mode background `--color-bg: #f6f2ea`:

| Value | Hex | Contrast |
|---|---|---|
| Old (pre-PR) | `#81879b` | 3.20:1 ❌ (below AA 4.5:1) |
| **Actual new** | `#5a637c` | **5.36:1** ✅ |
| **PR body claims** | `#616882` | 4.94:1 ✅ |

Both pass WCAG AA, so the fix is fine, but the PR description is **factually inconsistent** with the committed diff: it cites `#616882` (4.94:1) while the file ships `#5a637c` (5.36:1). Likely an auto-generated body that wasn't re-synced after a final color tweak. Dark mode `#a7b2cf` on `#0e141f` = 8.71:1, matching the body's claim.

**Minor polish:** 5.36:1 still falls short of the AA **Large Text** threshold's *AA* band (3:1) only by virtue of being small text — verify `text-faint` is never used for body copy under ~18pt or ~14pt bold. A spot check against `globals.css` consumers would close the loop.

### Nits

- `src/app/globals.css:1-16` — the `--color-text-faint` value now sits between `--color-text` (`#121723`, ~16:1) and `--color-text-muted` (`#5e6375`). Worth a visual check that the three-step ladder still reads as distinct rather than collapsed; an eye-test screenshot would be ideal.
- The B4 test docstring cites "lines 220-241 in `page.tsx`" for the `SyntaxHighlighter` mapping — that citation will rot. Replace with a code reference that survives refactors (e.g. anchor on the exported `WikiMarkdown` symbol, once extracted).

### Suggested path

1. Extract `sanitizeSchema` + plugin arrays + `<WikiMarkdown>` into a shared module.
2. Update both `page.tsx` and `MarkdownPreviewTabs.tsx` to import the shared module.
3. Update the test to import the same module and render via `<WikiMarkdown>` through `renderToStaticMarkup`.
4. Tighten the four assertions noted above (`language-js` regex, `javascript:` regex, `<script>` content, and either enable `rehype-raw` for the onclick test or rename it).
5. Fix the PR body to match the committed color, or commit `#616882` if that was the intended value.

After step 1–3, the B4 tests become a real regression guard instead of a parallel-implementation test.

Issue #273, opened 2026-07-08
Stack: Add math (KaTeX) + Mermaid to markdown pipeline


SweetSophia:
## Source
REVIEW.md §B.5 librarian, Week 2

## Problem
No math rendering or diagram support in the markdown pipeline.

## Suggested fix
- `remark-math` + `rehype-katex` for LaTeX math
- `remark-mermaid` for flowcharts/sequence diagrams
- Tiny bundles, big perceived-product wins

SweetSophia:
Implemented in PR #275 (squashed as `88a4126`). KaTeX math rendering + Mermaid diagrams are live. Deployed and verified: topics=49 articles=240 apiKeys=15.

PR #275, opened 2026-07-09
feat: KaTeX math + Mermaid diagrams in markdown pipeline (#273)

SweetSophia:
## Pull Request Summary

This PR extends the wiki markdown rendering pipeline to support **LaTeX math** (via KaTeX) and **Mermaid diagrams**, while keeping the existing sanitizer protections intact.

### Key Changes

**1. LaTeX Math via KaTeX**
- `remark-math` + `rehype-katex` are added to the rehype pipeline so inline `$... and block `$...$` syntax is parsed and rendered to KaTeX HTML + MathML.
- The rehype chain order ensures KaTeX runs **before** `rehype-sanitize`, so the rendered math output flows through the sanitizer rather than bypassing it.
- The sanitize schema was extended with KaTeX MathML tags (`annotation`, `math`, `mfrac`, `mi`, `mn`, `mo`, `mrow`, `msqrt`, `msub`, `msup`, `semantics`, etc.), plus needed attributes (`encoding`, `xmlns` on `math`).
- `span` was loosened to allow `aria-hidden`, `className`, and `style` so KaTeX visual rendering, accessibility hints, and styling survive sanitization.

**2. Mermaid Diagrams**
- `MermaidDiagram` is a new `"use client"` component that **dynamically imports** mermaid only after hydration (avoiding SSR browser-global issues).
- The markdown `code` component override intercepts fenced ` ```mermaid ` blocks and routes them to the diagram component instead of the syntax highlighter.
- A `pre` component override inspects the child code element's `language-mermaid` marker and **unwraps the parent `<pre>`**, producing a single `<figure class="mermaid-container">` in SSR output rather than the invalid `<pre><pre>…</pre></pre>` nesting.
- `securityLevel: "strict"` is set on every mermaid render to prevent inline scripts/handlers in the generated SVG.
- Until hydration completes, a `<pre class="mermaid-fallback">` displays the diagram source so SSR markup remains readable and no content flashes.

**3. Sanitizer Behavior Preserved**
- No `rehype-raw` was added — raw HTML in markdown body continues to be rendered as inert text.
- KaTeX's default `trust: false` blocks `\href{javascript:…}`, `\htmlStyle{…}`, `\htmlClass{…}`, etc., and any generated `style`/`href` attributes are also blocked by the sanitizer.
- Mermaid `strict` mode plus client-only execution keeps XSS surface minimal.

**4. Styling & Dependency Surface**
- KaTeX stylesheet imported globally in `globals.css`.
- New utility classes (`.katex-display`, `.mermaid-container`, `.mermaid-rendered`, `.mermaid-fallback`) provide scrollable blocks for wide math/equations and diagram rendering.
- Adds runtime deps for `katex`, `mermaid`, `remark-math`, and `rehype-katex`.

### Test Coverage Added

Five new security-oriented tests in `markdown-sanitize.test.ts`:
- Inline + block KaTeX rendering produces `katex` class, `aria-hidden="true"`, `<math>`, and `<annotation encoding="application/x-tex">` with no `<script>` output.
- `\href{javascript:…}` math payloads are scrubbed to text and never emit `href="javascript:…"` or `<script>` tags.
- KaTeX HTML extensions (`\htmlStyle`/\htmlClass`) cannot inject user-controlled `style`/`class`/`href` attributes.
- Mermaid `language-mermaid` blocks produce the `.mermaid-container` wrapper, expose the diagram source for client hydration, omit the `language-mermaid` class, and—critically—do **not** wrap the `<figure>` inside a `<pre>`.
- Malicious mermaid source containing `<script>` is escaped by the fallback renderer and never reaches the DOM as executable HTML.

The existing B4 sanitizer tests were tightened to also assert KaTeX tag allowlist (`math`) and the new `span` attributes (`ariaHidden`, `className`, `style`).

---

This pull request adds KaTeX math rendering and Mermaid diagram support to the wiki markdown pipeline (resolving #273). Two issues were fixed in the integration:

- **Mermaid initialization race condition (W1)**: The `MermaidDiagram` component now uses a module-level guard flag (`mermaidInitialized`) to ensure Mermaid is initialized exactly once per page load. Previously, multiple diagrams on the same page could race to call `mermaid.initialize()` and clobber each other's configuration.

- **Unnecessary re-renders (W2)**: The diagram source string is now memoized via `useMemo`, since `.trim()` returned a fresh string on every render and would cause the rendering effect to re-fire unnecessarily when parent components re-rendered.

- **Sanitize schema cleanup**: Removed the redundant `aria-hidden` attribute from the wiki markdown sanitize schema (the camelCase `ariaHidden` variant is sufficient for HTMLHint compatibility).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,6 +17,8 @@
         "isomorphic-dompurify": "^3.15.0",
         "js-yaml": "^4.1.1",
         "jszip": "^3.10.1",
+        "katex": "^0.17.0",
+        "mermaid": "^11.16.0",
         "next": "16.2.6",
         "next-auth": "^4.24.14",
         "pg": "^8.20.0",
@@ -26,8 +28,10 @@
         "react-markdown": "^10.1.0",
         "react-syntax-highlighter": "^16.1.1",
         "rehype-highlight": "^7.0.2",
+        "rehype-katex": "^7.0.1",
         "rehype-sanitize": "^6.0.0",
-        "remark-gfm": "^4.0.1"
+        "remark-gfm": "^4.0.1",
+        "remark-math": "^6.0.0"
       },
       "devDependencies": {
         "@testing-library/react": "^16.3.2",
@@ -43,6 +47,19 @@
         "happy-dom": "^20.10.6",
         "tsx": "^4.22.4",
         "typescript": "^5"
+      }
+    },
+    "node_modules/@antfu/install-pkg": {
+      "version": "1.1.0",
+      "resolved": "https://registry.npmjs.org/@antfu/install-pkg/-/install-pkg-1.1.0.tgz",
+      "integrity": "sha512-MGQsmw10ZyI+EJo45CdSER4zEb+p31LpDAFp2Z3gkSd1yqVZGi0Ebx++YTEMonJy4oChEMLsxZ64j8FH6sSqtQ==",
+      "license": "MIT",
+      "dependencies": {
+        "package-manager-detector": "^1.3.0",
+        "tinyexec": "^1.0.1"
+      },
+      "funding": {
+        "url": "https://github.com/sponsors/antfu"
       }
     },
     "node_modules/@asamuzakjp/css-color": {
@@ -341,6 +358,12 @@
         "node": ">=6.9.0"
       }
     },
+    "node_modules/@braintree/sanitize-url": {
+      "version": "7.1.2",
+      "resolved": "https://registry.npmjs.org/@braintree/sanitize-url/-/sanitize-url-7.1.2.tgz",
+      "integrity": "sha512-jigsZK+sMF/cuiB7sERuo9V7N9jx+dhmHHnQyDSVdpZwVutaBu7WvNYqMDLSgFgfB30n452TP3vjDAvFC973mA==",
+      "license": "MIT"
+    },
     "node_modules/@bramus/specificity": {
       "version": "2.4.2",
       "resolved": "https://registry.npmjs.org/@bramus/specificity/-/specificity-2.4.2.tgz",
@@ -352,6 +375,12 @@
       "bin": {
         "specificity": "bin/cli.js"
       }
+    },
+    "node_modules/@chevrotain/types": {
+      "version": "11.1.2",
+      "resolved": "https://registry.npmjs.org/@chevrotain/types/-/types-11.1.2.tgz",
+      "integrity": "sha512-U+HFai5+zmJCkK86QsaJtoITlboZHBqrVketcO2ROv865xfCMSFpELQoz1GkX5GzME8pTa+3kbKrZHQtI0gdbw==",
+      "license": "Apache-2.0"
     },
     "node_modules/@csstools/color-helpers": {
       "version": "6.0.2",
@@ -1214,6 +1243,23 @@
         "url": "https://github.com/sponsors/nzakas"
       }
     },
+    "node_modules/@iconify/types": {
+      "version": "2.0.0",
+      "resolved": "https://registry.npmjs.org/@iconify/types/-/types-2.0.0.tgz",
+      "integrity": "sha512-+wluvCrRhXrhyOmRDJ3q8mux9JkKy5SJ/v8ol2tu4FVjyYvtEzkc/3pK15ET6RKg4b4w4BmTk1+gsCUhf21Ykg==",
+      "license": "MIT"
+    },
+    "node_modules/@iconify/utils": {
+      "version": "3.1.4",
+      "resolved": "https://registry.npmjs.org/@iconify/utils/-/utils-3.1.4.tgz",
+      "integrity": "sha512-b1S7B1k9ohZ+iNTi2ATxbRYG9fTrJmUT0rc46bvVnNxqNRGW7dyo/vRREwyniI5IRN2RSJHDcm+s3BjWrSAjHw==",
+      "license": "MIT",
+      "dependencies": {
+        "@antfu/install-pkg": "^1.1.0",
+        "@iconify/types": "^2.0.0",
+        "import-meta-resolve": "^4.2.0"
+      }
+    },
     "node_modules/@img/colour": {
       "version": "1.1.0",
       "resolved": "https://registry.npmjs.org/@img/colour/-/colour-1.1.0.tgz",
@@ -1741,6 +1787,15 @@
       "resolved": "https://registry.npmjs.org/@kurkle/color/-/color-0.3.4.tgz",
       "integrity": "sha512-M5UknZPHRu3DEDWoipU6sE8PdkZ6Z/S+v4dD+Ke8IaNlpdSQah50lz1KtcFBa2vsdOnwbbnxJwVM4wty6udA5w==",
       "license": "MIT"
+    },
+    "node_modules/@mermaid-js/parser": {
+      "version": "1.2.0",
+      "resolved": "https://registry.npmjs.org/@mermaid-js/parser/-/parser-1.2.0.tgz",
+      "integrity": "sha512-oYPyv8A4As1yH5Bx+04iQEQxXuIQDe0GKCNSRgao6z8AM9jixXIfP0vsppRLvGf+nKIOb9/LdpWA4YuJiVvESA==",
+      "license": "MIT",
+      "dependencies": {
+        "@chevrotain/types": "~11.1.2"
+      }
     },
     "node_modules/@napi-rs/wasm-runtime": {
       "version": "0.2.12",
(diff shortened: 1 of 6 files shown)


SweetSophia:
## PR #275 Review — KaTeX math + Mermaid diagrams

**Scope:** `WikiMarkdown.tsx`, `MermaidDiagram.tsx`, `globals.css`, `markdown-sanitize.test.ts`, deps.
**Method:** Read diff + ran `npm run test:security` (34/34 pass) + traced `react-markdown`/`hast-util-to-jsx-runtime` behavior to verify the `pre` unwrap works as claimed + cross-checked against existing bot reviews (sourcery-ai, kody-ai, kilo-code).

### Summary

Architecture is sound: **KaTeX before sanitize in the rehype chain** (correct ordering), **Mermaid is client-only with `securityLevel: "strict"`**, **SSR fallback prevents layout shift**, and **5 new security tests cover the XSS surface**. I verified the `pre`-unwrap works because `hast-util-to-jsx-runtime` preserves the original hast `className` on the React element passed to the parent `pre` override (confirmed by tracing `element()` → `createChildren()` → `state.create()` and instrumenting the live render).

Two **production-affecting** issues, one **dead-code-defending test**, and a few suggestions.

---

### Warnings (address before merge)

**W1 — `mermaid.initialize()` called on every effect run** (`MermaidDiagram.tsx:36`)
The effect calls `mermaid.initialize({...})` on every mount/dependency flip. This (a) re-sets global mermaid config on every diagram and (b) lets two diagrams on the same page race to clobber each other's theme. Move initialization to a module-level singleton or `useRef` guard so it runs exactly once per page. Aligns with **SSOT** (Single Source of Truth) for mermaid config.

**W2 — `source = code.trim()` produces a new string per render** (`MermaidDiagram.tsx:64`)
`source` is in the effect's dependency array. Since `.trim()` returns a fresh string on every render, the effect flips on every parent re-render → re-imports mermaid, re-initializes, re-renders. Stabilize via `useMemo(() => code.trim(), [code])` (or drop `.trim()` from the dep-tracking path and only use it inside the effect).

**W3 — Test locks in a redundant `aria-hidden` allowlist entry** (`WikiMarkdown.tsx:51-52`, `markdown-sanitize.test.ts:40`)
```ts
span: [...(defaultSchema.attributes?.span ?? []), "aria-hidden", "ariaHidden", "className", "style"]
…
assert.ok(ARTICLE_SANITIZE_SCHEMA.attributes?.span?.includes("ariaHidden"));
```
`rehype-sanitize` normalizes HTML attributes to React-prop names before schema lookup, so `ariaHidden` (camelCase) is the only key that matters. The `"aria-hidden"` entry is dead allowlist surface (**YAGNI**). Remove the kebab-case entry and the corresponding assertion so future sanitizers can drop the duplication without breaking the test.

---

### Suggestions

**S1 — `pre` unwrap logic is fragile by design** (`WikiMarkdown.tsx:73-83`)
The comment says *"the `code` child is still an unrendered React element, so we inspect its props.className"*. That works today because `hast-util-to-jsx-runtime` propagates the original `className` onto the React element regardless of whether the `code` override is a component or string. But this is an implementation detail of react-markdown. A more robust approach: set the `pre` component on the hast tree directly (e.g. a small remark/rehype visitor that mutates `<pre><code class="language-mermaid">…` to `<figure data-mermaid>` before rendering), or use `unwrapDisallowed`.

**S2 — `dangerouslySetInnerHTML` with mermaid SVG is acceptable but worth a CVE check** (`MermaidDiagram.tsx:64`)
`securityLevel: "strict"` disables click handlers, `javascript:` URLs, and HTML in labels — good. But `dangerouslySetInnerHTML` is a standing red flag. Confirm mermaid `11.16.0` is current against the latest advisories (the historical `mermaid` CVE for arbitrary file read via diagrams was fixed in `11.4.0`; `11.16.0` should be clean, but worth a one-line sanity check pinned in the PR description or a `package.json` comment).

**S3 — No error boundary around `MermaidDiagram`**
The effect's try/catch handles mermaid parse errors, but a failure in `await import("mermaid")` (e.g. chunk-load failure on a flaky network) would propagate. A parent error boundary keeps a broken diagram from taking down the article page.

**S4 — `KATEX_TAG_NAMES` is a hardcoded subset** (`WikiMarkdown.tsx:12-38`)
If KaTeX adds tags in a future minor (it has done so between `0.16` and `0.17`), they silently won't render and won't be covered by tests. A comment pointing to the KaTeX version the list was sourced from would help future maintainers, or pull the list from a KaTeX export if one exists.

**S5 — `cancelled` flag prevents state updates but doesn't unwind mermaid's side effects**
When the component unmounts mid-`mermaid.render()`, mermaid may leave a transient DOM node. The `cancelled` flag only guards `setRenderState`. If mermaid exposes a cancellation/abort hook, use it; otherwise accept the minor leak and document it.

---

### Security analysis

| Vector | Mitigation | Status |
|---|---|---|
| KaTeX `\href{javascript:…}` | KaTeX `trust: false` strips + sanitize blocks `href` on allowed tags | ✅ covered by test `#273: math links cannot render javascript hrefs` |
| KaTeX `\htmlStyle`/`\htmlClass` injection | KaTeX `trust: false` + sanitize strips `style`/`class` user content | ✅ covered by test `#273: KaTeX HTML extensions…` |
| Raw `<script>` in markdown body | No `rehype-raw` → raw HTML treated as text; sanitize strips `<script>` | ✅ covered by existing B4 tests |
| Mermaid `<script>` in diagram source | Source rendered as React text children in fallback (auto-escaped) | ✅ covered by test `#273: mermaid fallback escapes…` |
| Mermaid SVG XSS via `dangerouslySetInnerHTML` | `securityLevel: "strict"` disables click handlers, `javascript:` URLs, HTML in labels | ✅ mitigated; see S2 re: CVE check |
| Sanitizer regression (e.g. someone removes `rehype-sanitize`) | Test `B4: shared markdown config preserves syntax-highlight and math classes` asserts the schema is present in `ARTICLE_REHYPE_PLUGINS` | ✅ good guardrail |

**STRIDE quick pass:** Tampering (XSS) — mitigated; Repudiation — N/A; Information disclosure — mermaid `strict` mode prevents exfil via crafted diagrams; Denial of service — large mermaid diagrams could slow client render (consider a soft cap on diagram size); Elevation of privilege — N/A.

---

### Architecture notes

- **Rehype ordering** `[rehypeKatex, [rehypeSanitize, ARTICLE_SANITIZE_SCHEMA]]` is correct — KaTeX output flows through sanitize, not around it.
- **Sanitizer schema** correctly extends `defaultSchema` with KaTeX MathML tags + needed attributes; `span` allowlist matches KaTeX's emitted attributes.
- **Dynamic import** of mermaid (`await import("mermaid")`) avoids SSR browser-global issues — correct pattern.
- **`useId` sanitization** (`replace(/[^A-Za-z0-9_-]/g, "")`) handles React 18's `:r0:` style IDs — correct.

### Test coverage assessment

The 5 new tests are well-targeted: positive rendering (KaTeX markup, mermaid wrapper), negative XSS (`javascript:` href, `htmlStyle`/`htmlClass`, mermaid `<script>`), and the structural assertion (no `<pre><figure>` nesting). The `B4` config test was tightened to assert the new schema entries. **One gap:** no test asserts `mermaid.initialize` is called exactly once, or that the effect doesn't re-run on parent re-render — adding one would have caught W1/W2.

---

### Verdict

**Approve with changes recommended.** W1 and W2 are production-affecting (first page with two mermaid diagrams, or first re-render of an article view). W3 is a minor cleanup. S1–S5 are optional.

> Cross-references: kilo-code-bot independently flagged W1, W2, W3 in its review at https://github.com/SweetSophia/noosphere/pull/275#issuecomment-4927485493 — findings are consistent.

SweetSophia:
Pushed `f2ca359` addressing W1–W3:

**W1** — `mermaid.initialize()` moved to a module-level `mermaidInitialized` guard so it runs exactly once per page load, even with concurrent diagrams.

**W2** — `source` now wrapped in `useMemo(() => code.trim(), [code])` so the effect only re-fires when the actual `code` prop changes, not on every parent re-render.

**W3** — Removed the redundant `"aria-hidden"` (kebab-case) entry from `ARTICLE_SANITIZE_SCHEMA.attributes.span`. Only `"ariaHidden"` (the hast property name that `rehype-sanitize` normalizes to) is kept.

All gates green: `tsc --noEmit` clean, `eslint` 0 errors, `test:security` 13/13, `next build` success.

SweetSophia:
## Final Review — PR #275 (post-`f2ca359`)

Re-checked after the fix commit. **All three warnings resolved cleanly.** 34/34 security tests pass; CI green across Analyze, CodeQL, lint-and-typecheck, docker, and all three bot reviews.

### Fix verification

| ID | Fix | Location | Verdict |
|---|---|---|---|
| **W1** | `mermaidInitialized` module-level guard; `mermaid.initialize(...)` only runs once per page load | `MermaidDiagram.tsx:14-16, 42-53` | ✅ Correct — solves the multi-diagram race; no need for `useRef` since mermaid config is genuinely process-global |
| **W2** | `useMemo(() => code.trim(), [code])` stabilizes the dep | `MermaidDiagram.tsx:21-23` | ✅ Correct — effect now only re-runs when `code` content actually changes |
| **W3** | `"aria-hidden"` removed from span allowlist | `WikiMarkdown.tsx:49-54` | ✅ Correct — `ariaHidden` (camelCase) is the only key `rehype-sanitize` reads; test assertion at `markdown-sanitize.test.ts:40` still validates the correct key |

### Remaining suggestions (non-blocking, from previous round)

- **S1** `pre` unwrap still relies on `hast-util-to-jsx-runtime` propagating `className` onto the React element — works today, still implementation-detail fragile. Acceptable for now; revisit if react-markdown major-bumps.
- **S2** `dangerouslySetInnerHTML` with mermaid SVG — `securityLevel: "strict"` is the right mitigation. No action needed unless mermaid advisories change.
- **S3** No error boundary around `MermaidDiagram` — still a nice-to-have for chunk-load failures, not a blocker.
- **S4** `KATEX_TAG_NAMES` hardcoded subset — acceptable; add a `// source: KaTeX 0.17` comment if you want future-proofing.
- **S5** `cancelled` flag doesn't unwind mermaid side effects — minor leak, document if it becomes a problem.

### Verdict

**Approve.** Ready to merge.

PR #277, opened 2026-07-10
feat: rate limiter follow-up enhancements (timeout, Retry-After, oversized header short-circuit, EVALSHA)

SweetSophia:
### Summary

This PR layers four follow-up hardening enhancements onto the existing Redis-backed rate limiter: safe header handling before hashing, accurate Retry-After signaling, bounded eval latency, and optimized script delivery. Together they close specific issues identified during review of the preceding rate-limit work (PR #276).

#### 1. Oversized proxy-header identifier short-circuit (`src/lib/rate-limit.ts`)
- Adds a `MAX_IDENTIFIER_LENGTH = 256` constant and bails early with `"invalid:oversized"` when an untrusted proxy header exceeds it.
- Previously, oversized values flowed straight into SHA-256 hashing, wasting CPU on attacker-controlled kilobyte-scale inputs and still allowing those values to influence the fallback map key shape.
- The hash path is preserved for legitimate-length but malformed identifiers.

#### 2. `Retry-After` header on 429 responses (`src/lib/rate-limit.ts`)
- `rateLimitExceeded` now accepts `windowMs` and sets `Retry-After: ceil(windowMs / 1000)` on the 429 response, per RFC 7231 §7.1.3.
- `degradedResult` and every 429-returning branch in `rateLimit` propagate `options.windowMs` so the header reflects the actual configured window rather than a hardcoded value.
- Ceiling division guarantees the wait hint is never smaller than the true window.

#### 3. Eval timeout via `Promise.race` (`src/lib/rate-limit-redis.ts`)
- Adds `EVAL_TIMEOUT_MS = 1500` covering the case where ioredis reports `ready: true` yet a command stalls mid-flight (e.g., wedged socket, network partition).
- New `timeoutRace` helper races the eval against a `setTimeout` rejection; the timer is always cleared in `finally`, avoiding leaks.
- On timeout, an explicit error is thrown so the existing catch block in `rateLimit` routes through `degradedResult`, preserving the fail-closed degradation semantics.
- The `checkRedisRateLimit` signature widens to require both `eval` and `evalsha` on the client.

#### 4. `EVALSHA` with automatic `EVAL` fallback (`src/lib/rate-limit-redis.ts`)
- Precomputes `SCRIPT_SHA` once via `createHash("sha1").update(REDIS_RATE_LIMIT_SCRIPT).digest("hex")`.
- New `evalWithFallback` helper tries `evalsha` first; on a `NOSCRIPT` error (script evicted from Redis cache) it transparently retries with full `eval`, mirroring and slightly extending ioredis's built-in pattern with explicit error matching.
- Eliminates re-sending the ~400-byte script body on every call and keeps behavior deterministic when the cache misses.
- Test-helper `FakeRedisClient` computes the matching SHA, accepts `evalsha` calls, delegates to shared `runRateLimitScript` logic, and still emits `NOSCRIPT` for unknown SHAs so the fallback path is exercisable.

#### 5. Test coverage updates
- **`fake-redis.ts`**: Adds `evalshaCalls` / `evalFallbackCalls` counters, a static `SCRIPT_SHA` matching production, an `evalsha` implementation with NOSCRIPT simulation, and extracts shared `runRateLimitScript` so both `eval` and `evalsha` exercise identical logic.
- **`rate-limit.test.ts`**: Three new tests
  - Oversized (512-byte) identifier short-circuits to `invalid:oversized` and the key length stays compact.
  - 429 responses carry the correct `Retry-After: 30` value for a 30-second window.
  - `evalsha` is called on the hot path (counter assertion).
- **`rate-limit-fallback.test.ts`**: The Redis-recovery monkey-patch is updated from `eval` → `evalsha`, with restored fallback delegation to the original `evalsha`.

### Functional Impact

- **Security**: Resource-bound protection against oversized proxy-header abuse and a bounded blast radius for stuck Redis connections.
- **Compliance**: 429s now self-document their retry interval, improving client behavior under throttling.
- **Performance**: Per-request Redis payload is reduced to just the SHA and key arguments on the hot path; cache-miss recovery remains correct.
- **Reliability**: Wedged-socket scenarios no longer block request handling indefinitely; they fall through to the existing degraded path with the same observability (a `Redis eval timed out after 1500ms` warning).

---

## Summary

This pull request enhances the Redis rate limiter with several follow-up improvements referenced in issue #242.

### Script Export for Testing Consistency
- Exported `REDIS_RATE_LIMIT_SCRIPT` as a public constant so test helpers can compute the SHA1 hash from the same source string. This eliminates the previous duplication where the test fixture hard-coded the script body, which created a risk of the two staying out of sync.

### Script Hash Deduplication
- Replaced the duplicated hash literal in the test helper (`fake-redis.ts`) with one derived from the exported constant, ensuring any future edits to the Lua script body are automatically reflected in tests.

### Cleanup of Unused Tracking Field
- Removed the now-unused `evalFallbackCalls` counter from the fake Redis client.

### Robust Timeout Handling
- Improved `timeoutRace()` to silently swallow late rejections from the losing Redis promise once the race has settled. Previously, if the timeout fired first, the underlying `evalsha`/`eval` promise could still reject afterward (e.g. broken pipe, cluster failover) and surface as an unhandled promise rejection. The catch handler is attached only after the race resolves, so the first rejection still propagates correctly to the caller when it wins.

## Impact
- **Testing**: Eliminates a class of silent test bugs caused by script hash drift between production and test code.
- **Reliability**: Prevents late Redis failures from producing noisy unhandled promise rejection warnings while still surfacing real errors to the degraded fallback path when they are timely.

---

## Summary

This PR enhances the Redis rate limiter with several follow-up improvements.

### Test coverage: EVALSHA with NOSCRIPT fallback

The existing test for EVALSHA was expanded to cover the full fallback path. Previously it only verified that EVALSHA was called first; now it also forces a NOSCRIPT error by overriding `evalsha` to always throw, then confirms that the EVAL fallback path is exercised and the rate limit decision still succeeds (returns `{ allowed: true }`). This locks in the NOSCRIPT recovery behavior so future refactors don't silently regress it.

---

## Summary

This PR introduces follow-up enhancements to the Redis rate limiter, addressing three reliability hazards (H1/H2) discovered in the previous review.

### Changes

**`src/lib/rate-limit-redis.ts` — Timeout race hardening (H1)**
- `checkRedisRateLimit` now accepts a `disconnect()`-capable Redis client. When the EVAL/EVALSHA call hits its timeout, the client is forcibly disconnected before the function rejects. This prevents the wedged-socket problem where a Redis instance reports `status="ready"` but silently deadlocks subsequent commands by queueing them onto a non-responsive socket. The next call will get a freshly rebuilt connection.
- `evalWithFallback` now takes an `aborted` context shared with the timeout race. If the outer race has already settled via timeout, no follow-up `EVAL` will fire on a NOSCRIPT response (H2: prevents counter drift from late writes against a connection the caller has already abandoned).
- `timeoutRace` accepts an `onTimeout` callback invoked exactly once when the timeout wins the race. Earlier it could fire on any loser rejection; now it only fires when timeout actually triggered.

**`src/lib/rate-limit.ts` — Oversized-header short-circuit**
- `x-forwarded-for` is sliced to 1 024 bytes **before** splitting/trimming. A multi-kilobyte header no longer wastes CPU on string parsing. The cap is generous for any plausible proxy chain. Other headers (`x-real-ip`, `cf-connecting-ip`) already have natural upper bounds from HTTP layer limits and remain uncapped.

### Tests

**`src/__tests__/rate-limit.test.ts`**
- The EVAL fallback (NOSCRIPT) test now asserts that `evalKeys.length` increased by exactly one and that the new key contains `keyPrefix`. Previously, the test only verified `allowed: true`, which would have passed even if the fallback silently no-op'd (H3 — silent degradation masquerading as success).
- The recovery test now calls `await fakeRedis.connect()` on the freshly swapped-in `FakeRedisClient` so its `ready` state is observable in the same way as the initial setup.

### Issue Reference

Closes the follow-up items from #242.

---

## Summary

This PR enhances the Redis rate limiter with several follow-up improvements focused on timeout handling, reliability, and graceful degradation.

### Key Changes

1. **Configurable timeout** — `checkRedisRateLimit` now accepts an optional `timeoutMs` parameter, defaulting to the existing internal `EVAL_TIMEOUT_MS`. This allows callers (and tests) to tune the timeout per use case.

2. **Disconnect-once guarantee** — Replaced the direct `redis.disconnect()` call with a new `disconnectRedisClientOnce(redis)` helper. Previously, concurrent timed-out checks could each call `disconnect()` independently, potentially causing redundant or racy disconnects. The helper ensures the client is disconnected only once even under concurrency, while remaining safe against repeated invocations from the singleton recovery path.

3. **Late NOSCRIPT and late EVAL handling** — When the timeout fires during a Redis call:
   - If a NOSCRIPT error surfaces **after** the timeout, the rate limiter no longer falls back to `EVAL` (which would issue an additional round-trip on an already-degraded client). The late error is rethrown so the request stays on the fail-closed path.
   - If an `EVAL` call returns **after** the timeout, its result is discarded. Since the command has already been executed in Redis, the side effect cannot be rolled back, so the late response is thrown away and the limiter remains fail-closed.

4. **Test infrastructure improvements** — Extended `FakeRedisClient` to simulate:
   - Configurable delays on `EVALSHA` and `EVAL` calls
   - Optional `NOSCRIPT` error injection from `EVALSHA`
   - Optional short-circuit for `EVAL` results after a delay
   - A `disconnectKeepsStatus` option that mirrors ioredis's real behavior of staying `"ready"` until the socket emits close, enabling accurate testing of the disconnect-once guard.

5. **New test coverage** — Added three tests verifying:
   - A wedged (slow) Redis client is disconnected exactly once when concurrent checks time out, even though the client status remains `"ready"`.
   - No `EVAL` is issued when a NOSCRIPT error arrives after the timeout.
   - A late but successful `EVAL` response is discarded, and the disconnect-once guard still triggers exactly once.

---

This pull request builds on the existing Redis rate limiter with several follow-up enhancements focused on robustness around timeouts, oversized header handling, and Lua script caching (EVALSHA).

Key changes:

- **Timeout-induced disconnect safety**: Replaces the prior `disconnectRedisClientOnce` call with `invalidateRedisClient`, and adds a regression test confirming that when cleanup throws during a timeout, the original "Redis eval timed out" errors are still preserved for concurrent callers and the singleton is removed (so a reconnect is suppressed during cooldown). This prevents a wedged but status="ready" socket from silently queuing commands.

- **Late EVAL response handling**: Clarifies the comment around discarding late EVAL responses — the request falls back to the warmed per-process limiter rather than the degraded fail-closed path, accepting the small risk of one Redis slot being consumed even if the local decision differs, since other instances depend on the shared counter.

- **Fake Redis test helper improvements**: Renames `evalResultAfterDelay` to `evalReturnValue` for clarity, and adds support for simulating a `disconnect()` that throws an error, enabling the new timeout/disconnect-error regression test.

- **Test refactor**: Imports `getReadyRedisClient` to assert the singleton was actually invalidated, and renames a test to better describe its behavior ("discards a late EVAL response that settles after the timeout").

Note: The pull request title also references "Retry-After" and "oversized header short-circuit" enhancements, but the provided diff only shows changes related to timeout/EVAL behavior and test helper improvements. Those additional features may be in other files not included in this patch set.

---

**feat: rate limiter follow-up enhancements (timeout, Retry-After, oversized header short-circuit, EVALSHA)**

This PR hardens the Redis-backed rate limiter around four follow-up scenarios surfaced after the initial rate-limiter work.

**Timeout cleanup ordering** — `timeoutRace` now runs the cleanup callback synchronously inside the timer before rejecting the race. This prevents any queued NOSCRIPT handlers or subsequent callers from observing the abandoned client as still usable, and wraps cleanup errors so they cannot mask the original timeout error reported to callers.

**Retry-After header** — Sets the `Retry-After` response header on 429 responses (window seconds, ceiled) so well-behaved clients back off correctly. (Also corrects a minor comment typo.)

**Oversized proxy identifier handling** — Replaces the previous hashing path with an explicit length bound so attacker-controlled `x-forwarded-for` values cannot produce arbitrarily large rate-limit keys. Covered by a regression test using a 12,000-character identifier.

**EVALSHA-first execution** — The atomic check now uses `EVALSHA` (with `EVAL` fallback for NOSCRIPT), which is verified by updated tests asserting the SHA-based call path is exercised on both invalid-response and thrown-error fallback paths.

**Singleton recovery and reconnect guard fixes** —
- After a timeout, the wedged singleton is rebuilt after the reconnect cooldown; an integration test asserts that a new client is constructed and serves subsequent requests.
- Disconnect cleanup is allowed to retry after a synchronous failure (no permanent poison of the once-guard).
- A stale failed connect attempt can no longer evict a replacement singleton that was installed concurrently.
- When both connect and disconnect throw, the singleton is still cleared and the reconnect cooldown prevents the same failed client from being handed out again.

**Test infrastructure** — `FakeRedisClient.connect()` now sets `status = "connecting"` before awaiting the simulated delay so concurrent-connection tests mirror real `ioredis` behavior.

### DIFF
--- a/src/__tests__/_helpers/fake-redis.ts
+++ b/src/__tests__/_helpers/fake-redis.ts
@@ -1,4 +1,5 @@
 import { createHash } from "node:crypto";
+import { REDIS_RATE_LIMIT_SCRIPT } from "@/lib/rate-limit-redis";
 
 type PipelineCommand =
   | { method: "zremrangebyscore"; args: [string, string, string] }
@@ -16,7 +17,6 @@ export class FakeRedisClient {
   connectCalls = 0;
   readonly evalKeys: string[] = [];
   evalshaCalls = 0;
-  evalFallbackCalls = 0;
 
   private readonly values = new Map<string, string>();
   private readonly expires = new Map<string, number>();
@@ -94,11 +94,8 @@ export class FakeRedisClient {
     return 1;
   }
 
-  /** SHA1 of the rate-limit script body.  Must match rate-limit-redis.ts. */
   private static readonly SCRIPT_SHA =
-    createHash("sha1")
-      .update('\nredis.call("ZREMRANGEBYSCORE", KEYS[1], "-inf", ARGV[1])\nlocal count = redis.call("ZCARD", KEYS[1])\n\nif count >= tonumber(ARGV[3]) then\n  return 0\nend\n\nredis.call("ZADD", KEYS[1], ARGV[2], ARGV[4])\nredis.call("EXPIRE", KEYS[1], ARGV[5])\nreturn 1\n')
-      .digest("hex");
+    createHash("sha1").update(REDIS_RATE_LIMIT_SCRIPT).digest("hex");
 
   async evalsha(
     sha1: string,
--- a/src/__tests__/_helpers/fake-redis.ts
+++ b/src/__tests__/_helpers/fake-redis.ts
@@ -1,3 +1,5 @@
+import { createHash } from "node:crypto";
+
 type PipelineCommand =
   | { method: "zremrangebyscore"; args: [string, string, string] }
   | { method: "zcard"; args: [string] }
@@ -13,6 +15,8 @@ export class FakeRedisClient {
   status = "wait";
   connectCalls = 0;
   readonly evalKeys: string[] = [];
+  evalshaCalls = 0;
+  evalFallbackCalls = 0;
 
   private readonly values = new Map<string, string>();
   private readonly expires = new Map<string, number>();
@@ -90,6 +94,27 @@ export class FakeRedisClient {
     return 1;
   }
 
+  /** SHA1 of the rate-limit script body.  Must match rate-limit-redis.ts. */
+  private static readonly SCRIPT_SHA =
+    createHash("sha1")
+      .update('\nredis.call("ZREMRANGEBYSCORE", KEYS[1], "-inf", ARGV[1])\nlocal count = redis.call("ZCARD", KEYS[1])\n\nif count >= tonumber(ARGV[3]) then\n  return 0\nend\n\nredis.call("ZADD", KEYS[1], ARGV[2], ARGV[4])\nredis.call("EXPIRE", KEYS[1], ARGV[5])\nreturn 1\n')
+      .digest("hex");
+
+  async evalsha(
+    sha1: string,
+    numberOfKeys: number,
+    ...args: Array<string | number>
+  ): Promise<number> {
+    this.assertReady();
+    this.evalshaCalls++;
+    if (sha1 !== FakeRedisClient.SCRIPT_SHA) {
+      throw new Error("NOSCRIPT No matching script. Please use EVAL.");
+    }
+    // Delegate to the existing eval logic (skip the script body arg).
+    this.evalKeys.push(args[0] as string);
+    return this.runRateLimitScript(numberOfKeys, args);
+  }
+
   async eval(
     _script: string,
     numberOfKeys: number,
@@ -100,11 +125,26 @@ export class FakeRedisClient {
       throw new Error("FakeRedisClient only supports one-key rate-limit scripts");
     }
 
+    // Validate basic types (runRateLimitScript does full validation).
+    if (typeof args[0] !== "string" || typeof args[4] !== "string") {
+      throw new Error("Invalid rate-limit script arguments");
+    }
+    this.evalKeys.push(args[0] as string);
+    return this.runRateLimitScript(numberOfKeys, args);
+  }
+
+  private runRateLimitScript(
+    numberOfKeys: number,
+    args: Array<string | number>
+  ): number {
+    if (numberOfKeys !== 1) {
+      throw new Error("FakeRedisClient only supports one-key rate-limit scripts");
+    }
+
     const [key, windowStart, now, maxRequests, member, ttlSeconds] = args;
     if (typeof key !== "string" || typeof member !== "string") {
       throw new Error("Invalid rate-limit script arguments");
     }
-    this.evalKeys.push(key);
 
     this.zremrangebyscore(key, "-inf", String(windowStart));
     const count = this.zcard(key);
--- a/src/__tests__/_helpers/fake-redis.ts
+++ b/src/__tests__/_helpers/fake-redis.ts
@@ -10,23 +10,40 @@ type PipelineCommand =
 interface FakeRedisClientOptions {
   connectDelayMs?: number;
   rejectConnect?: boolean;
+  evalshaDelayMs?: number;
+  evalshaError?: Error;
+  evalDelayMs?: number;
+  evalResultAfterDelay?: number;
+  disconnectKeepsStatus?: boolean;
 }
 
 export class FakeRedisClient {
   status = "wait";
   connectCalls = 0;
+  disconnectCalls = 0;
   readonly evalKeys: string[] = [];
   evalshaCalls = 0;
+  evalCalls = 0;
 
   private readonly values = new Map<string, string>();
   private readonly expires = new Map<string, number>();
   private readonly sortedSets = new Map<string, Array<{ score: number; member: string }>>();
   private readonly connectDelayMs: number;
   private readonly rejectConnect: boolean;
+  private readonly evalshaDelayMs: number;
+  private readonly evalshaError: Error | undefined;
+  private readonly evalDelayMs: number;
+  private readonly evalResultAfterDelay: number | undefined;
+  private readonly disconnectKeepsStatus: boolean;
 
   constructor(options: FakeRedisClientOptions = {}) {
     this.connectDelayMs = options.connectDelayMs ?? 0;
     this.rejectConnect = options.rejectConnect ?? false;
+    this.evalshaDelayMs = options.evalshaDelayMs ?? 0;
+    this.evalshaError = options.evalshaError;
+    this.evalDelayMs = options.evalDelayMs ?? 0;
+    this.evalResultAfterDelay = options.evalResultAfterDelay;
+    this.disconnectKeepsStatus = options.disconnectKeepsStatus ?? false;
   }
 
   async connect() {
@@ -41,7 +58,10 @@ export class FakeRedisClient {
   }
 
   disconnect() {
-    this.status = "end";
+    this.disconnectCalls += 1;
+    if (!this.disconnectKeepsStatus) {
+      this.status = "end";
+    }
   }
 
   async get(key: string): Promise<string | null> {
@@ -104,6 +124,11 @@ export class FakeRedisClient {
   ): Promise<number> {
     this.assertReady();
     this.evalshaCalls++;
+    await delay(this.evalshaDelayMs);
+    if (this.evalshaError) {
+      throw this.evalshaError;
+    }
+    this.assertReady();
     if (sha1 !== FakeRedisClient.SCRIPT_SHA) {
       throw new Error("NOSCRIPT No matching script. Please use EVAL.");
     }
@@ -117,6 +142,13 @@ export class FakeRedisClient {
     numberOfKeys: number,
     ...args: Array<string | number>
   ): Promise<number> {
+    this.assertReady();
+    this.evalCalls += 1;
+    await delay(this.evalDelayMs);
+    if (this.evalResultAfterDelay !== undefined) {
+      this.evalKeys.push(args[0] as string);
+      return this.evalResultAfterDelay;
+    }
     this.assertReady();
     if (numberOfKeys !== 1) {
       throw new Error("FakeRedisClient only supports one-key rate-limit scripts");
@@ -200,5 +232,11 @@ export class FakeRedisClient {
     if (this.status !== "ready") {
       throw new Error("Redis command executed before client was ready");
     }
+  }
+}
+
+async function delay(ms: number): Promise<void> {
+  if (ms > 0) {
+    await new Promise((resolve) => setTimeout(resolve, ms));
   }
 }
--- a/src/__tests__/_helpers/fake-redis.ts
+++ b/src/__tests__/_helpers/fake-redis.ts
@@ -51,6 +51,9 @@ export class FakeRedisClient {
 
   async connect() {
     this.connectCalls += 1;
+    // ioredis leaves "wait" synchronously when connect() starts. Preserve
+    // that transition so concurrent-connection tests exercise real behavior.
+    this.status = "connecting";
     if (this.connectDelayMs > 0) {
       await new Promise((resolve) => setTimeout(resolve, this.connectDelayMs));
     }
--- a/src/__tests__/_helpers/fake-redis.ts
+++ b/src/__tests__/_helpers/fake-redis.ts
@@ -13,8 +13,9 @@ interface FakeRedisClientOptions {
   evalshaDelayMs?: number;
   evalshaError?: Error;
   evalDelayMs?: number;
-  evalResultAfterDelay?: number;
+  evalReturnValue?: number;
   disconnectKeepsStatus?: boolean;
+  disconnectError?: Error;
 }
 
 export class FakeRedisClient {
@@ -33,8 +34,9 @@ export class FakeRedisClient {
   private readonly evalshaDelayMs: number;
   private readonly evalshaError: Error | undefined;
   private readonly evalDelayMs: number;
-  private readonly evalResultAfterDelay: number | undefined;
+  private readonly evalReturnValue: number | undefined;
   private readonly disconnectKeepsStatus: boolean;
+  private readonly disconnectError: Error | undefined;
 
   constructor(options: FakeRedisClientOptions = {}) {
     this.connectDelayMs = options.connectDelayMs ?? 0;
@@ -42,8 +44,9 @@ export class FakeRedisClient {
     this.evalshaDelayMs = options.evalshaDelayMs ?? 0;
     this.evalshaError = options.evalshaError;
     this.evalDelayMs = options.evalDelayMs ?? 0;
-    this.evalResultAfterDelay = options.evalResultAfterDelay;
+    this.evalReturnValue = options.evalReturnValue;
     this.disconnectKeepsStatus = options.disconnectKeepsStatus ?? false;
+    this.disconnectError = options.disconnectError;
   }
 
   async connect() {
@@ -59,6 +62,9 @@ export class FakeRedisClient {
 
   disconnect() {
     this.disconnectCalls += 1;
+    if (this.disconnectError) {
+      throw this.disconnectError;
+    }
     if (!this.disconnectKeepsStatus) {
       this.status = "end";
     }
@@ -145,9 +151,9 @@ export class FakeRedisClient {
     this.assertReady();
     this.evalCalls += 1;
     await delay(this.evalDelayMs);
-    if (this.evalResultAfterDelay !== undefined) {
+    if (this.evalReturnValue !== undefined) {
       this.evalKeys.push(args[0] as string);
-      return this.evalResultAfterDelay;
+      return this.evalReturnValue;
     }
     this.assertReady();
     if (numberOfKeys !== 1) {
--- a/src/__tests__/rate-limit-fallback.test.ts
+++ b/src/__tests__/rate-limit-fallback.test.ts
@@ -207,12 +207,16 @@ describe("Rate Limiter Fallback (Redis failures)", () => {
   });
 
   it("warns again when Redis recovers and a later degradation begins", async () => {
-    const originalEval = fakeRedis.eval.bind(fakeRedis);
+    const originalEvalsha = fakeRedis.evalsha.bind(fakeRedis);
     let shouldFail = true;
-    fakeRedis.eval = async (...args: Parameters<FakeRedisClient["eval"]>) => {
+    fakeRedis.evalsha = (async (
+      sha1: string,
+      numberOfKeys: number,
+      ...args: Array<string | number>
+    ) => {
       if (shouldFail) throw new Error("Redis unavailable");
-      return originalEval(...args);
-    };
+      return originalEvalsha(sha1, numberOfKeys, ...args);
+    }) as typeof fakeRedis.evalsha;
 
     const request = makeRequest("10.0.0.63");
     const options = { windowMs: 60_000, maxRequests: 10, keyPrefix: "warning-transition" };
--- a/src/__tests__/rate-limit-fallback.test.ts
+++ b/src/__tests__/rate-limit-fallback.test.ts
@@ -153,13 +153,18 @@ describe("Rate Limiter Fallback (Redis failures)", () => {
   });
 
   it("falls back to in-process limiter when the atomic check returns invalid data", async () => {
-    fakeRedis.eval = async () => null as never;
+    let evalshaCalled = false;
+    fakeRedis.evalsha = async () => {
+      evalshaCalled = true;
+      return null as never;
+    };
 
     const request = makeRequest("10.0.0.60");
     const options = { windowMs: 60_000, maxRequests: 1, keyPrefix: "pipe-null" };
 
     // First request: invalid Redis response → warmed fallback allows.
     assert.deepStrictEqual(await rateLimit(request, options), { allowed: true });
+    assert.equal(evalshaCalled, true);
 
     // Second request: fallback blocks (bucket full).
     const result2 = await rateLimit(request, options);
@@ -167,7 +172,9 @@ describe("Rate Limiter Fallback (Redis failures)", () => {
   });
 
   it("falls back to in-process limiter when the atomic check throws", async () => {
-    fakeRedis.eval = async () => {
+    let evalshaCalled = false;
+    fakeRedis.evalsha = async () => {
+      evalshaCalled = true;
       throw new Error("Redis connection lost");
     };
 
@@ -181,6 +188,7 @@ describe("Rate Limiter Fallback (Redis failures)", () => {
     try {
       // First request: Redis throws → warmed fallback allows.
       assert.deepStrictEqual(await rateLimit(request, options), { allowed: true });
+      assert.equal(evalshaCalled, true);
 
       // Second request: fallback blocks (bucket full).
       const result2 = await rateLimit(request, options);
--- a/src/__tests__/rate-limit-redis-integration.test.ts
+++ b/src/__tests__/rate-limit-redis-integration.test.ts
@@ -2,6 +2,11 @@ import { after, before, describe, it } from "node:test";
 import assert from "assert";
 import Redis from "ioredis";
 import { checkRedisRateLimit } from "@/lib/rate-limit-redis";
+import {
+  getReadyRedisClient,
+  _redisTestHooks,
+} from "@/lib/cache/redis";
+import { FakeRedisClient } from "./_helpers/fake-redis";
 
 const redisUrl = process.env.REDIS_URL;
 
@@ -69,5 +74,57 @@ describe("Redis rate limiter integration", { skip: !redisUrl }, () => {
       }),
       true
     );
+  });
+
+  it("rebuilds the timed-out singleton after the reconnect cooldown", async () => {
+    const originalDateNow = Date.now;
+    let currentTime = originalDateNow();
+    const wedgedClient = new FakeRedisClient({
+      evalshaDelayMs: 40,
+      disconnectKeepsStatus: true,
+    });
+    await wedgedClient.connect();
+    _redisTestHooks.setClientForTesting(wedgedClient as never);
+    Date.now = () => currentTime;
+
+    try {
+      await assert.rejects(
+        checkRedisRateLimit(wedgedClient as never, {
+          key: `test:ratelimit:recovery-timeout:${crypto.randomUUID()}`,
+          windowMs: 60_000,
+          maxRequests: 5,
+          now: currentTime,
+          timeoutMs: 25,
+        }),
+        /Redis eval timed out/
+      );
+
+      assert.equal(await getReadyRedisClient(), null);
+      currentTime += 5_001;
+      // getReadyRedisClient checks the mocked clock synchronously before its
+      // first await. Restore the real clock immediately afterward so ioredis
+      // internals do not inherit the test clock while connecting.
+      const recoveryPromise = getReadyRedisClient();
+      Date.now = originalDateNow;
+      const recoveredClient = await recoveryPromise;
+      assert.ok(recoveredClient);
+      assert.notStrictEqual(recoveredClient, wedgedClient);
+
+      const key = `test:ratelimit:recovered:${crypto.randomUUID()}`;
+      keys.add(key);
+      assert.equal(
+        await checkRedisRateLimit(recoveredClient, {
+          key,
+          windowMs: 60_000,
+          maxRequests: 5,
+          now: Date.now(),
+          member: "after-recovery",
+        }),
+        true
+      );
+    } finally {
+      Date.now = originalDateNow;
+      _redisTestHooks.reset();
+    }
   });
 });
--- a/src/__tests__/rate-limit.test.ts
+++ b/src/__tests__/rate-limit.test.ts
@@ -110,6 +110,48 @@ describe("Redis Rate Limiter", () => {
     assert.ok(fakeRedis.evalKeys[0].length < 128);
   });
 
+  it("short-circuits oversized identifiers without hashing", async () => {
+    const oversized = "x".repeat(512);
+    const request = makeRequest(oversized);
+
+    assert.deepStrictEqual(
+      await rateLimit(request, {
+        windowMs: 60_000,
+        maxRequests: 5,
+        keyPrefix: "oversized-identifier",
+      }),
+      { allowed: true }
+    );
+
+    // The key should use the "invalid:oversized" short-circuit, not a hash.
+    assert.equal(fakeRedis.evalKeys.length, 1);
+    assert.ok(fakeRedis.evalKeys[0].includes("invalid:oversized"));
+    assert.ok(fakeRedis.evalKeys[0].length < 80);
+  });
+
+  it("includes a Retry-After header on 429 responses", async () => {
+    const request = makeRequest("10.0.0.20");
+    const options = { windowMs: 30_000, maxRequests: 1, keyPrefix: "retry-after" };
+
+    await rateLimit(request, options); // First request OK
+    const result = await rateLimit(request, options); // Second blocked
+
+    assert.equal(result.allowed, false);
+    if (!result.allowed) {
+      assert.equal(result.response.status, 429);
+      assert.equal(result.response.headers.get("Retry-After"), "30");
+    }
+  });
+
+  it("uses EVALSHA and falls back to EVAL on NOSCRIPT", async () => {
+    // The FakeRedisClient.evalsha throws NOSCRIPT for unknown SHAs.
+    // Verify that a normal call uses evalsha first.
+    const request = makeRequest("10.0.0.21");
+    await rateLimit(request, { windowMs: 60_000, maxRequests: 5, keyPrefix: "evalsha-test" });
+
+    assert.ok(fakeRedis.evalshaCalls > 0, "EVALSHA should be called first");
+  });
+
   it("falls back to in-process limiter when Redis client is unavailable", async () => {
     // Simulate Redis not configured by passing null-like state
     const request = makeRequest("10.0.0.6");
--- a/src/__tests__/rate-limit.test.ts
+++ b/src/__tests__/rate-limit.test.ts
@@ -6,7 +6,7 @@ import {
   fallbackLimiter,
   _fallbackLimiterTestHooks,
 } from "@/lib/rate-limit-fallback";
-import { _redisTestHooks } from "@/lib/cache/redis";
+import { getReadyRedisClient, _redisTestHooks } from "@/lib/cache/redis";
 import { NextRequest } from "next/server";
 import { FakeRedisClient } from "./_helpers/fake-redis";
 
@@ -205,9 +205,46 @@ describe("Redis Rate Limiter", () => {
 
     assert.equal(results[0].status, "rejected");
     assert.equal(results[1].status, "rejected");
-    // Exercise the singleton recovery cleanup while status still reads ready.
-    _redisTestHooks.reset();
     assert.equal(fakeRedis.disconnectCalls, 1);
+    // The singleton is removed synchronously and the cooldown suppresses an
+    // immediate reconnect even though the fake still reports status="ready".
+    assert.equal(await getReadyRedisClient(), null);
+  });
+
+  it("preserves concurrent timeout errors when disconnect cleanup throws", async () => {
+    fakeRedis = new FakeRedisClient({
+      evalshaDelayMs: 40,
+      disconnectError: new Error("disconnect failed"),
+    });
+    await fakeRedis.connect();
+    _redisTestHooks.setClientForTesting(fakeRedis as never);
+    const originalConsoleError = console.error;
+    console.error = () => {};
+
+    try {
+      const options = {
+        key: "ratelimit:disconnect-error:10.0.0.26",
+        windowMs: 60_000,
+        maxRequests: 5,
+        now: Date.now(),
+        timeoutMs: 25,
+      };
+      const results = await Promise.allSettled([
+        checkRedisRateLimit(fakeRedis as never, options),
+        checkRedisRateLimit(fakeRedis as never, options),
+      ]);
+
+      for (const result of results) {
+        assert.equal(result.status, "rejected");
+        if (result.status === "rejected") {
+          assert.match(String(result.reason), /Redis eval timed out/);
+        }
+      }
+      assert.equal(fakeRedis.disconnectCalls, 1);
+      assert.equal(await getReadyRedisClient(), null);
+    } finally {
+      console.error = originalConsoleError;
+    }
   });
 
   it("does not issue EVAL when NOSCRIPT arrives after the timeout", async () => {
@@ -231,11 +268,11 @@ describe("Redis Rate Limiter", () => {
     assert.equal(fakeRedis.evalKeys.length, 0);
   });
 
-  it("discards a successful EVAL response that settles after the timeout", async () => {
+  it("discards a late EVAL response that settles after the timeout", async () => {
     fakeRedis = new FakeRedisClient({
       evalshaError: new Error("NOSCRIPT No matching script. Please use EVAL."),
       evalDelayMs: 40,
-      evalResultAfterDelay: 0,
+      evalReturnValue: 0,
     });
     await fakeRedis.connect();
 
--- a/src/__tests__/rate-limit.test.ts
+++ b/src/__tests__/rate-limit.test.ts
@@ -1,6 +1,7 @@
 import { describe, it, beforeEach, afterEach } from "node:test";
 import assert from "assert";
 import { rateLimit, _rateLimitTestHooks } from "@/lib/rate-limit";
+import { checkRedisRateLimit } from "@/lib/rate-limit-redis";
 import {
   fallbackLimiter,
   _fallbackLimiterTestHooks,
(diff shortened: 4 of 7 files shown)


SweetSophia:
## In-depth re-review

**Verdict:** Approve with revisions. Two high-impact issues (H1, H2) deserve a follow-up patch before merge; remaining items are medium-or-lower polish.

I re-read the current files, ran the rate-limit suite locally (35/35 pass, typecheck clean, only the pre-existing unrelated `descendants` lint warning), and reconciled against bot review comments already on the thread. Two ioredis-specific concerns flagged by kilo-code-bot (NOSCRIPT regex; `err.code` property) were verified against `node_modules/ioredis@5.10.1` source — they are **not** real issues (ioredis's `ReplyError` has no `.code`; its own Script.js uses `err.message.indexOf('NOSCRIPT')` per `node_modules/ioredis/built/Script.js:19,49`). The implementation is correct on that axis. However, two distinct timeout-side concerns ARE real.

---

### H1 — Wedged `ready` Redis client is never disconnected after EVAL timeout (HIGH)

**File:** `src/lib/rate-limit-redis.ts:66-68` (call site) and `src/lib/cache/redis.ts:73-99` (caller).

**Evidence (ioredis 5.10.1 source):**
- `node_modules/ioredis/built/Redis.js:341-343` — per-command timeout is only enforced if `commandTimeout` is set; otherwise no timeout.
- `node_modules/ioredis/built/Redis.js:408` — `commandQueue` and `:377` `offlineQueue` keep accepting writes from a `ready` client even with a wedged socket.
- `node_modules/ioredis/built/Redis.js:345` — `writable = this.status === "ready" || ...`, so each new request keeps queueing.

**Failure sequence:**
1. Redis hangs mid-command. Client `status` remains `ready`.
2. `Promise.race` rejects after 1500 ms. Caller returns degraded response. ✅
3. The in-flight `redis.evalsha(...)` promise is *still pending*; the socket never wakes.
4. Next request arrives. `getReadyRedisClient()` returns the same client (status is still `ready`). New `redis.evalsha(...)` is queued onto the wedged socket. The queue grows by one per request.
5. After ~N requests / 90 s, default `maxRetriesPerRequest: 20` doesn't apply because the client never reconnects; eventually the socket sees TCP RST or the user's `retryStrategy` returns `null` (`cache/redis.ts:50-54` rejects after first retry), but **only on the next failure event** — not proactively.

This is the exact scenario the timeout was meant to fix, and the patch only fixes half of it. ChatGPT Codex P2 already flagged this.

**Fix sketch (≤10 lines):**

```ts
// In timeoutRace, on timeout, return a sentinel + caller resets the client:
import { redisClient } from "@/lib/cache/redis";   // pass via opts

// On timeoutRace timeout:
client.disconnect();  // forces next getReadyRedisClient() to rebuild
```

Or, equivalently, set `commandTimeout: EVAL_TIMEOUT_MS` on the ioredis client so ioredis itself rejects the in-flight command at the same boundary; the late rejection then doesn't match NOSCRIPT (Q4 confirmation: the regex `/NOSCRIPT/i` does not match `"Command timed out"`), so the EVAL fallback would be skipped.

---

### H2 — `evalWithFallback` can issue EVAL after the outer race has settled (HIGH)

**File:** `src/lib/rate-limit-redis.ts:81-94`.

**Confirmed by both oracle and librarian lanes** (sourced against ioredis source + Redis 7.4 docs).

**Failure sequence:**
1. HTTP handler awaits `timeoutRace(evalPromise, 1500)`.
2. Timer fires → race rejects → outer await throws → `rateLimit`'s `catch` flips to `degradedResult`; client response sent.
3. Underlying `redis.evalsha` is still pending. Redis is healthy but the script cache was evicted (Redis 7.4+ evicts EVAL-loaded scripts when the cache fills — see Redis docs "EVALSHA").
4. Eventually Redis responds `NOSCRIPT No matching script. Please use EVAL.`
5. `evalWithFallback`'s catch matches → `await redis.eval(REDIS_RATE_LIMIT_SCRIPT, 1, ...args)` against the *same* Redis server. This Lua script atomically runs `ZREMRANGEBYSCORE / ZCARD / ZADD / EXPIRE` and writes to Redis **after** the HTTP layer has already moved on.
6. Result: counter drift between in-process state and Redis state. Two app instances can observe inconsistent windows for up to one full `windowMs` after a single NOSCRIPT race.

**Cluster amplification:** `ClusterOptions.d.ts:54-77` — default `maxRedirections: 16` × default 100 ms retry delays = up to 1.6 s, which exceeds `EVAL_TIMEOUT_MS = 1500`. Failover can deliberately trigger the race.

**Fix sketch (≤6 lines):**

```ts
async function checkRedisRateLimit(redis, options) {
  const ctx = { aborted: false };
  const evalPromise = evalWithFallback(redis, args, ctx);
  const result = await timeoutRace(evalPromise, EVAL_TIMEOUT_MS, options.key, () => {
    ctx.aborted = true;
  });
  ...
}

async function evalWithFallback(redis, args, ctx) {
  try {
    return await redis.evalsha(SCRIPT_SHA, 1, ...args);
  } catch (err) {
    if (ctx?.aborted) throw err;  // do NOT issue EVAL after outer race settled
    if (/NOSCRIPT/i.test(err.message)) {
      return await redis.eval(REDIS_RATE_LIMIT_SCRIPT, 1, ...args);
    }
    throw err;
  }
}
```

This is exactly what Gemini's bot suggestion `5` and `6` proposed (medium priority); my verification confirms it should be uplifted to high priority.

---

### M1 — Phase-2 NOSCRIPT test only asserts return value, not that EVAL ran

**File:** `src/__tests__/rate-limit.test.ts:154-167`.

```ts
fakeRedis.evalsha = async () => {
  throw new Error("NOSCRIPT No matching script. Please use EVAL.");
};
try {
  const result = await rateLimit(makeRequest("10.0.0.22"), { ... });
  assert.deepStrictEqual(result, { allowed: true });
}
```

If the NOSCRIPT detection regex ever breaks (e.g., tightened to `err.message.startsWith("NOSCRIPT ")` and Redis changes its reply), this test still passes via `degradedResult` returning `{ allowed: true }` for a fresh key's `'allowed'` fallback (`rate-limit.ts:82-91, 143-146`). The EVAL fallback path is never proven to have run.

**Fix sketch:**

```ts
const evalshaBefore = fakeRedis.evalshaCalls;
try {
  const result = await rateLimit(makeRequest("10.0.0.22"), { ... });
  assert.deepStrictEqual(result, { allowed: true });
  // EVAL fallback must have run — overlay replaced the original, so only eval()
  // could have produced the second evalKeys entry.
  assert.equal(fakeRedis.evalshaCalls, evalshaBefore);     // shadowed, no counter
  assert.equal(fakeRedis.evalKeys.length, 2);              // phase 1 + phase 2
  assert.ok(fakeRedis.evalKeys[1].includes("10.0.0.22"));
} finally {
  fakeRedis.evalsha = originalEvalsha;
}
```

Captures both the precise failure mode Codex P2 highlighted and the silent regression kody-bot warned about (NOSCRIPT detection breaking → still green).

---

### M2 — `Retry-After` always reports the full window

**File:** `src/lib/rate-limit.ts:78`.

`String(Math.ceil(windowMs / 1000))` is an upper bound, not the actual remaining time. A 60 s window where the user's most-recent request was 50 s ago gets told to wait 60 s. RFC 7231 / RFC 6585 allow delta-seconds; behavior is conformant but pessimizes throughput for clients that strictly honor the header.

Severity: minor. If you want a precise value, you'd need to consult the local fallback bucket's earliest timestamp (`fallbackLimiter.entries.get(key).timestamps[0]`). I'd lean toward leaving the upper bound and adding a code comment, since precise values are misleading during in-progress windows and the upper bound is the safe default for clients.

---

### M3 — Doc comment lies about the implementation

**File:** `src/lib/rate-limit-redis.ts:43`.

> "Uses EVALSHA with automatic EVAL fallback (via ioredis defineCommand pattern) to avoid re-sending the script body on every call."

`defineCommand` is **not** used. The hand-rolled `evalWithFallback` is. Per librarian verification, ioredis's own `Script.js` (which `defineCommand` uses internally) ALSO does message-regex NOSCRIPT detection — switching would not improve correctness, only remove the drift risk. The cheaper fix is to fix the comment.

**Fix sketch:** replace "(via ioredis defineCommand pattern)" with "(manual evalsha/eval fallback on NOSCRIPT, matching ioredis's built-in behavior)".

---

### L1 — `getClientIdentifier` materializes full `x-forwarded-for` before length cap

**File:** `src/lib/rate-limit.ts:21-26`.

For a 12 KB `x-forwarded-for`, `.split(",")[0]?.trim()` walks the entire string. `MAX_IDENTIFIER_LENGTH = 256` does protect the SHA-256 path, but the parse work still runs. Trivial to cap earlier:

```ts
const raw = (request.headers.get("x-real-ip")
  ?? request.headers.get("cf-connecting-ip")
  ?? request.headers.get("x-forwarded-for")?.split(",")[0]?.trim()
  ?? "unknown").slice(0, 1024).trim();
```

Severity: nit. `x-forwarded-for` is normally a few hundred bytes. Worth doing only if you ever see a 12 KB header in the wild.

---

### L2 — `pipeline()` in FakeRedisClient still uses pre-PR script semantics

**File:** `src/__tests__/_helpers/fake-redis.ts:157-197`.

The pipeline methods are dead in production now (single-shot EVALSHA replaces the 4-command sequence). Not a bug; could carry a comment that pipeline support is for legacy rollback-test sequencing, or delete it if no test uses it. Worth a grep before deletion.

---

## Verified strengths

- **S1** — `Promise.race` + `finally.catch(() => {})` is correctly placed. Q3 of the ioredis-research lane confirmed against ECMAScript `sec-promise.race` and V8 unhandled-rejection semantics: the in-place catch installs a handler before the late rejection's microtask checkpoint, and only AFTER the race settles so a real first rejection still propagates. (`rate-limit-redis.ts:116-123`)
- **S2** — `/NOSCRIPT/i.test(err.message)` is the documented ioredis pattern. The `ReplyError` constructor (`node_modules/redis-errors/lib/modern.js:31-38`) does NOT set `.code`. ioredis's own test fixture (`test/functional/scripting.ts:249-255`) asserts on the message text. The kilocode `err.code === "NOSCRIPT"` suggestion would silently never match.
- **S3** — Exported `REDIS_RATE_LIMIT_SCRIPT` with derived `SCRIPT_SHA` (commit `f60fbbd`) eliminates test/production drift.
- **S4** — The actual NOSCRIPT fallback test (commit `4110725`) now monkey-patches `evalsha` to throw NOSCRIPT — this is the correct shape; only the assertion precision is missing (M1).
- **S5** — TTL_GRACE_SECONDS rationale is now accurate (carry-over from PR #276).
- **S6** — Existing fallback capacity-saturation test (PR #276) correctly asserts `entries.size === capacity` after denial, locking in the fail-closed semantics.

---

## Local verification

```
$ npm run test:rate-limit    # 35/35 pass (suite "Rate Limiter Edge Cases" + friends)
$ npm run typecheck          # clean
$ npm run lint               # 1 pre-existing unrelated warning (admin/topics/actions.ts:99)
```

The PR is in a clean, mergeable state modulo H1 and H2; M1, M3 are quick wins that can land in the same PR. L1, L2 are nits.

---
*Reconciliation sources: PR #277 diff and 24 review-thread comments; ioredis 5.10.1 local source at `node_modules/ioredis/built/{Redis.js, Script.js, Command.js, redis/RedisOptions.d.ts}`; Redis 7.4+ EVAL script eviction docs; local PR worktree at `/tmp/noosphere-pr277` (commit 4110725).*

SweetSophia:
Thanks for the thorough review — all findings addressed in `19ec8c1`.

**H1 (wedged client)** ✅ — `timeoutRace` now calls `redis.disconnect()` on timeout. The `timedOut` flag ensures this only fires when the timeout actually wins the race, not in every `finally`. Next `getReadyRedisClient()` rebuilds the connection.

**H2 (late EVAL after race settled)** ✅ — Added abort context `{ aborted: boolean }` threaded through `evalWithFallback`. When the timeout fires, `ctx.aborted = true` is set before the EVAL fallback check, so a late NOSCRIPT error from an already-abandoned race will not issue a follow-up EVAL.

**M1 (NOSCRIPT test precision)** ✅ — Test now asserts `evalKeys.length` increased and the new key contains the expected `keyPrefix`, proving the EVAL path actually executed rather than silently passing via `degradedResult`.

**M3 (doc comment)** ✅ — Replaced "via ioredis defineCommand pattern" with "manual evalsha/eval fallback on NOSCRIPT, matching ioredis's built-in behaviour".

**L1 (oversized header CPU)** ✅ — `x-forwarded-for` now sliced to 1024 bytes before `.split(",")`, avoiding CPU waste on attacker-controlled kilobyte-scale headers.

**M2 (Retry-After precision)** — Left as-is per your recommendation (upper bound is the safe default). Existing comment already documents the rationale.

**L2 (dead pipeline in FakeRedis)** — Left in place; not worth the churn for a test helper.

**Bot findings not addressed** (verified non-issues per your analysis):
- kilo-code-bot `err.code === "NOSCRIPT"` — confirmed wrong (ioredis ReplyError has no `.code`)
- kilo-code-bot NOSCRIPT regex — confirmed correct (matches ioredis's own `Script.js` pattern)

37/37 tests pass. Typecheck + lint clean. Ready for merge.

SweetSophia:
## Second-pass re-review (`19ec8c1`)

**Verdict:** Approve with revisions. H1, H2, M1, M3, and L1 are correctly fixed and verified at the event-loop level. The remaining concerns are about **regression-test coverage** for the new timeout/disconnect sequencing and the **blast radius** of disconnecting a Redis singleton shared with the search-cache module.

### Confirmed fixes (verified against the V8 microtask spec and ioredis 5.10.1 source)

#### H1 — `redis.disconnect()` on timeout — Correct ✅
`src/lib/rate-limit-redis.ts:77-82`. Verified sequencing:

1. `setTimeout` rejects `timeoutPromise` (macrotask) → microtask `M1`.
2. `M1`: `Promise.race` rejects → `timeoutRace` `finally` runs synchronously: `clearTimeout`, `onTimeout()` (sets `ctx.aborted = true`, calls `redis.disconnect()`), `promise.catch(() => {})`.
3. Exception propagates to `rate-limit.ts:147` catch → degraded result.
4. Next `getRedisClient()` sees `status === "end"` → `resetRedisClient()` → rebuilds.

Wedged socket no longer persists across requests. The `timedOut` flag at `:396` ensures `onTimeout` only fires when the timer actually wins the race. No concurrency hazard: `disconnect()` is synchronous; idempotent on already-`"end"` clients.

#### H2 — Abort NOSCRIPT → EVAL after timeout — Correct ✅
`src/lib/rate-limit-redis.ts:73-115`. The `ctx = { aborted: false }` is fresh per call (no shared-state). The `ctx.aborted` check at `:108` is **strictly before** the NOSCRIPT regex at `:110`, so a post-timeout NOSCRIPT cannot trigger an EVAL. All three microtask orderings verified.

#### M1 — NOSCRIPT test assertion — Correct ✅
`src/__tests__/rate-limit.test.ts:155-176`. `evalKeysBefore` captures the Phase 1 length; the new `evalKeys.length === evalKeysBefore + 1` and `evalKeys[last].includes("evalsha-noscript")` assertions would have caught the original silent-fallback bug.

#### L1 — Header cap — Correct ✅
`src/lib/rate-limit.ts:20-29`. Behavior preserved for all normal inputs. Side benefit: an explicitly empty `x-forwarded-for` no longer pins the shared `"unknown"` bucket — minor improvement.

#### M3 — Inline docs — Covered
Six new comment blocks at `rate-limit-redis.ts:43-50, 70-72, 95-96, 106-107, 143` and one at `rate-limit.ts:22-23` document the abort/disconnect path. No CHANGELOG entry for these specific changes, but inline coverage is sufficient.

### New findings from this pass

#### N1 (Medium) — `redis.disconnect()` drops the singleton shared with `search-cache.ts` — operational blast radius beyond rate-limit

`getReadyRedisClient()` returns the **same** client instance used by `src/lib/cache/search-cache.ts` (lines 39, 53, 72, 96 — `redis.get` / `setex` / `incr`). When the rate-limit timeout fires and disconnects, every in-flight search-cache command rejects with `"Connection is closed."`

Pre-PR behavior: wedged socket → 1500ms stall on every Redis op (rate-limit + search-cache).
Post-PR behavior: rate-limit times out once → search-cache fails fast on every subsequent op until reconnect.

`search-cache.ts` has fail-open handling (`43-46, 60-63, 83-86, 100-103`), so no correctness regression — users get uncached results instead of crashes. **But two operational consequences worth flagging:**

1. **Behavior change**: a wedged Redis previously caused latency stalls; now it causes brief cache-effectiveness drops to 0% during reconnect.
2. **Reconnect storm risk**: see N6 below.

If the shared singleton was intentional pre-PR (likely — simpler ops), this is acceptable. If you want isolation, give rate-limit its own Redis instance. **This is the most important new finding from this pass** and worth discussing before merge.

#### N6 (Medium) — Post-disconnect reconnect has no cooldown — hot-path cost amplification

`REDIS_RECONNECT_COOLDOWN_MS = 5000` at `src/lib/cache/redis.ts:8` is only set when `client.connect()` throws (line 95), **not** when `redis.disconnect()` is called from inside the rate-limit timeout (line 81 of `rate-limit-redis.ts`).

Consequence: a single Redis timeout → every subsequent request does `new Redis()` + `client.connect()` with `connectTimeout: 2000`. If Redis is intermittently slow (one timeout per minute), every minute = full reconnect storm. Each request pays worst-case 2s timeout.

Recommended fix: in `resetRedisClient()` (`cache/redis.ts:11`), also set `nextConnectAttemptAt = Date.now() + REDIS_RECONNECT_COOLDOWN_MS` when the disconnect reason is timeout, not just on connect failure. Or apply the cooldown unconditionally in `resetRedisClient`.

#### N5 (Medium) — No regression tests for H1 or H2

`FakeRedisClient` lacks knobs to simulate a wedged socket. Recommended additions:
- `hangEvalshaMs?: number` — makes `evalsha` resolve only after this delay.
- `disconnectCalls: number` — counter incremented by `disconnect()`.

With these, two new tests would lock in the fix:
- **H1 test**: `hangEvalshaMs: 5000`; assert `rateLimit` resolves in ≤ `EVAL_TIMEOUT_MS + 50ms`; assert `disconnectCalls === 1`.
- **H2 test**: `hangEvalshaMs: 5000`; mid-timeout, have `evalsha` reject with NOSCRIPT; assert `eval` was **not** called.

Without these, future refactors can silently regress H1/H2.

#### N7 (Low) — `commandTimeout: 1500` constructor option is the ioredis-idiomatic primitive

`src/lib/cache/redis.ts:45-56`. ioredis 5.10.1 has a `commandTimeout` option that auto-rejects pending commands on slow connections without requiring `Promise.race`. Adopting it would:

- Make H1 belt-and-braces (ioredis handles the rejection itself).
- Eliminate the need for the manual `timeoutRace`/`ctx.aborted` mechanism entirely.

`Promise.race` works, but `commandTimeout` is one source of truth, not three (timeout + abort flag + finally catch). Worth a follow-up issue if the team wants to consolidate.

#### N2 (Low) — `x-real-ip` and `cf-connecting-ip` not size-capped

`src/lib/rate-limit.ts:25-26`. Only `x-forwarded-for` is capped at 1024 bytes. Framework caps headers at 16-32 KB, so not a real DoS, but cosmetic symmetry worth a one-liner.

#### N4 (Low) — `disconnect()` called twice during recovery transition

`src/lib/rate-limit-redis.ts:81` + `src/lib/cache/redis.ts:11-17`. After timeout: rate-limit calls `disconnect()`. Next request: `getRedisClient()` sees `"end"`, calls `resetRedisClient()`, which calls `disconnect()` again. Idempotent and harmless, but tightenable.

#### N3 (Low) — Redundant `await fakeRedis.connect()` in test

`src/__tests__/rate-limit.test.ts:329`. `setClientForTesting` is followed by explicit `await fakeRedis.connect()`. `getReadyRedisClient()` will drive the connect on next `rateLimit` anyway (line 88 of `cache/redis.ts`). Functionally identical; remove or add comment.

### Local verification

- `npm run test:rate-limit` → 35/35 pass + 2 integration tests hidden at the describe level (no local Redis; CI runs all 37 with Redis service container).
- `npm run typecheck` → clean.
- `npm run lint` → 1 pre-existing unrelated warning (`src/app/wiki/admin/topics/actions.ts:99`).
- CI: all 10 checks pass (lint, typecheck, docker, test, CodeQL, Analyze ×3, 3 bot reviews, Snyk).

### Recommended next steps (in priority order)

| # | Severity | Action |
|---|---|---|
| 1 | Medium | Decide on N1: keep shared singleton (document the tradeoff) or isolate rate-limit to its own Redis instance. |
| 2 | Medium | Add N6 cooldown in `resetRedisClient` — small change, prevents reconnect storm. |
| 3 | Medium | Add N5 regression tests for H1 and H2 — locks in the fix. |
| 4 | Low | N7 follow-up: consider migrating to `commandTimeout: 1500` to consolidate timeout handling. |
| 5 | Low | N2, N3, N4 cleanup. |

Nothing blocks merge; the H1/H2 fix is sound and addresses the wedged-socket class of bugs. N1 and N6 are the most important operational follow-ups.

SweetSophia:
## PR #277 review — head `19ec8c1`

Thanks for the iteration. The H1/H2/M1/M3/L1 fixes in `f60fbbd` + `19ec8c1` are correct on the paths exercised by existing tests, and the four enhancements (EVALSHA-with-fallback, EVAL timeout, oversized-identifier short-circuit, `Retry-After`) close the gaps from the PR #242 review cleanly. One substantive defect and one substantive coverage gap remain; the rest is operational follow-up.

---

### F1 (Medium) — `evalWithFallback` still has a counter-drift window after the timeout

**File**: `src/lib/rate-limit-redis.ts:101-115`

**Path**:
1. `redis.evalsha()` rejects with NOSCRIPT.
2. The `catch` enters the NOSCRIPT branch and `await redis.eval(...)` is queued against the live client.
3. While that `eval` is in flight, `timeoutRace`'s 1.5 s timer fires. Its `finally` runs `onTimeout()` → `ctx.aborted = true` and `redis.disconnect()`.
4. The outer `Promise.race` has already settled with the timeout error, so the value eventually returned by `redis.eval()` is silently dropped (and a late rejection is suppressed by `promise.catch(() => {})`).

The existing guard at line 108 (`if (ctx.aborted) throw err`) only protects the *pre-catch* race — i.e. the case where the timer wins before `evalsha` rejects. Once the catch has entered the NOSCRIPT branch and `await redis.eval()` is queued, `ctx.aborted` is never re-checked. If `redis.eval()` had already been pushed onto the wire before `disconnect()` took effect, Redis will still execute the script server-side, incrementing the ZSET on a counter the caller never sees — exactly the "late EVAL after the race has settled" class the abort context was added to prevent.

**Impact**: bounded and **fail-closed** — over-counting makes the user hit the limit *sooner*, never later, so this is not a rate-limit bypass. It is still a real correctness defect (counter/state drift between the Redis ZSET and the local fallback), and it's the reason kody-ai's race-condition flag survives this iteration.

**Minimal fix** (kody's suggestion, with one extra re-check inside the NOSCRIPT branch):

```ts
} catch (err: unknown) {
  if (ctx.aborted) throw err;                                   // pre-catch race
  if (err instanceof Error && /NOSCRIPT/i.test(err.message)) {
    if (ctx.aborted) throw err;                                 // re-check before issuing EVAL
    return (await redis.eval(REDIS_RATE_LIMIT_SCRIPT, 1, ...args)) as number | string;
  }
  throw err;
}
```

For full coverage (a timer firing *during* the `await redis.eval(...)`), wrap the eval in a `Promise.race` against an abort signal that resolves when `ctx.aborted` flips:

```ts
const evalP = redis.eval(REDIS_RATE_LIMIT_SCRIPT, 1, ...args);
const abortP = new Promise<never>((_, reject) => {
  const tick = setInterval(() => {
    if (ctx.aborted) { clearInterval(tick); reject(new Error("aborted")); }
  }, 25);
});
try { return await Promise.race([evalP, abortP]); }
finally { clearInterval(tick); /* + ctx.aborted rethrow */ }
```

### F2 (Medium) — No regression tests for H1 / H2

`FakeRedisClient` exposes `connectDelayMs` and `rejectConnect` but no `evalshaHangMs` or `evalshaDelayedNoscriptMs`. The disconnect-on-timeout and abort-during-NOSCRIPT paths are not reachable from any existing test, so a future refactor of `timeoutRace` or `evalWithFallback` can silently regress H1/H2 without breaking CI.

Recommended additions:

- `evalshaHangMs?: number` — makes `evalsha` resolve only after this delay (to drive a real timeout).
- `disconnectCalls: number` — counter incremented inside the existing `disconnect()` (also useful for F1 verification).
- Test A (H1): `evalshaHangMs: 5_000`; assert `rateLimit` resolves in `≤ EVAL_TIMEOUT_MS + 50 ms`, `fakeRedis.status === "end"`, `disconnectCalls === 1`.
- Test B (H2): `evalshaHangMs: 5_000`; mid-timeout, swap `evalsha` to throw NOSCRIPT; assert `fakeRedis.evalKeys.length` is unchanged and the caller sees a degraded result.

These would lock both invariants and would have caught the F1 window above.

---

### Confirmed / accepted (no action requested)

- **H1** — `timeoutRace`'s `timedOut` flag (`rate-limit-redis.ts:127`) ensures `onTimeout` fires only when the timer actually wins. `disconnect()` is synchronous and idempotent on ioredis's already-`"end"` clients. Sequencing verified against the V8 microtask model.
- **H2 (partial)** — see F1 for the unfixed window.
- **M1** — `evalKeysBefore + 1` and `keyPrefix.includes(...)` assertions in `rate-limit.test.ts:155-176` correctly prove the EVAL fallback path executed.
- **M2** — `Retry-After = ceil(windowMs / 1000)` is the safe upper bound; agreed with your reasoning.
- **M3** — inline docs now accurately describe the abort/disconnect sequence.
- **L1** — 1024-byte `xff.slice` plus the 256-byte short-circuit is sound.
- **L2** — dead pipeline methods in `FakeRedisClient` left in place; not worth churn.

### Non-blocking observations

Most of these restate your own N1–N7; raising them publicly so they're not lost between iterations:

1. **N1 — shared singleton blast radius.** `getReadyRedisClient()` is also used by `src/lib/cache/search-cache.ts` (`redis.get`, `setex`, `incr`). A timeout-driven `redis.disconnect()` rejects every concurrent search-cache op. `search-cache` is fail-open, so no correctness regression — but the failure mode shifts from "latency stall" to "brief 0% cache effectiveness until reconnect". One-paragraph decision in the PR description (keep-shared vs isolate-instance).
2. **N6 — no reconnect cooldown after a timeout-disconnect.** `REDIS_RECONNECT_COOLDOWN_MS = 5000` is only armed in `getReadyRedisClient`'s catch when `client.connect()` throws (`cache/redis.ts:95`). `resetRedisClient()` itself does not set `nextConnectAttemptAt`, so a timeout-driven disconnect lets the next request immediately rebuild with `connectTimeout: 2000`. One-liner fix in `resetRedisClient`.
3. **N7 — ioredis `commandTimeout: 1500`.** Would let ioredis reject stuck commands itself and eliminate the manual `Promise.race` + abort-context + finally-catch machinery. Out of scope here; good follow-up issue.
4. **N2 / N3 / N4** — cosmetic: cap `x-real-ip` / `cf-connecting-ip` for symmetry with the XFF cap, drop redundant `await fakeRedis.connect()` in the recovery test (`rate-limit.test.ts:329`), fold the double `disconnect()` path.

Additional small items I noticed:

- `evalshaCalls` is a plain `number` while `evalKeys` is `readonly` (`fake-redis.ts:18-19`). Style drift; either both readonly or both mutable.
- `result === "1"` / `result === "0"` is dead defensive code (Redis returns numbers), but harmless.
- The 512-byte oversized-identifier test sends `"x".repeat(512)` and the short-circuit constant is `MAX_IDENTIFIER_LENGTH = 256`. A near-boundary test (256 vs 257) would tighten the contract.
- `console.warn` at `rate-limit-redis.ts:86` logs the key, which for hashed identifiers is `invalid:<sha256>` — safe to log, just noting.

---

### Verdict

**Approve with revisions.** Address **F1** before merge (the kody-ai race-condition flag was not fully closed by `19ec8c1`; the suggested re-check is two lines and removes the counter-drift window). **F2** (regression tests for H1/H2) is the highest-leverage follow-up but does not block merge. Everything else is operational or cosmetic and can be tracked as follow-up issues.

Local verification I ran: read the head diff, walked `timeoutRace` / `evalWithFallback` / `getReadyRedisClient` microtask sequencing, cross-checked the 31 call sites of `getReadyRedisClient`, and confirmed the 35-test suite described in the PR description all pass at `19ec8c1`. CI is green on all 10 checks.

SweetSophia:
@SweetSophia Follow-up to [your review](https://github.com/SweetSophia/noosphere/pull/277#issuecomment-4951056261): the remaining timeout/recovery gaps are fixed in `bba30c3`.

- Timeout handling now sets the abort flag and invalidates the shared Redis client **before** rejecting the outer race, preventing any new NOSCRIPT→EVAL fallback after timeout. The singleton is removed synchronously, reconnects are cooled down for 5 seconds, concurrent real-ioredis connects share one promise, and stale connect settlements cannot evict a replacement client. Disconnect failures are best-effort and release the identity guard so recovery can retry.
- The unavoidable limitation remains: Redis cannot cancel a command already written to the socket. An in-flight EVAL may therefore finish and consume one slot after the caller timed out; its late result is discarded. A compensating `ZREM` would be unsafe because the local fail-safe path may have legitimately admitted that request, so cleanup could undercount the shared limiter.
- Regression coverage now exercises wedged timeouts, late NOSCRIPT suppression, late EVAL completion, throwing disconnect cleanup, authentic EVALSHA invalid/error fallback paths, real ioredis `wait→connecting` behavior, stale-vs-replacement connects, and natural post-cooldown recovery through live Redis.

Verification: cache 8/8; Redis-backed rate-limit suite 45/45 twice consecutively; typecheck clean; lint 0 errors (one pre-existing unrelated warning). The final GitHub test, Docker, CodeQL, Snyk, Kilo, and Kody gates are all green after the full 20-minute window. Two final sequential fresh-context reviews independently found no correctness, security, reliability, or material test-authenticity issues.

PR #277 remains open and unmerged for your review.

SweetSophia:
## PR #277 — second-pass review (head `bba30c3`)

**Verdict: Approve for merge.** The three new commits (`207dfcf`, `ede628f`, `bba30c3`) close the prior F1 race and add a robust timeout-recovery hardening layer. One MEDIUM observation worth tracking, no HIGH/CRITICAL findings across any lane (correctness, security, design).

### F1 status — closed

`evalWithFallback` now has three `ctx.aborted` guards (`src/lib/rate-limit-redis.ts:111`, `:117`, `:129`) plus the synchronous `onTimeout()` placement in `timeoutRace` (`:155–161`). Race trace:

- **Pre-catch abort** → check 1 throws the original error → no NOSCRIPT branch entry ✓
- **NOSCRIPT-branch abort** → check 2 throws before `redis.eval()` is queued ✓
- **In-flight EVAL** → client disconnected by `onTimeout`'s `invalidateRedisClient` (which synchronously nulls the singleton + sets cooldown BEFORE throwing) → EVAL promise rejects → check 3 throws ✓

The intentional late-EVAL comment at `:123–128` correctly acknowledges the bounded fail-closed counter drift; removing the EVAL on abort would be unsafe because other instances rely on the shared Redis ZSET. Verified acceptable.

### MEDIUM — no direct regression test for the F1 closure mechanism

**Files**: `src/__tests__/rate-limit-redis-integration.test.ts`, `src/__tests__/rate-limit.test.ts`

The recovery integration test (`:79–128`) covers H1 + H2 *outcome* (singleton invalidation, cooldown, new client), and the pre-existing test at `rate-limit.test.ts:486–505` ("does not issue EVAL when NOSCRIPT arrives after the timeout") exercises NOSCRIPT-after-timeout. But no test specifically asserts the **line 117** invariant: *inside* the NOSCRIPT branch, between the regex match and the `await redis.eval()`, abort must be re-checked.

A future refactor that drops the `:117` guard still passes today's suite — the catch-level `:111` check is sufficient as long as timeout fires before the catch block enters, which is the dominant case. The `:117` guard specifically protects the narrower window where the NOSCRIPT regex matches *after* the outer timer callback has fired but *before* the catch's microtask sees `ctx.aborted`. That's exactly the window where the kody-ai flag originated.

**Minimal addition (5 lines)**: a test where `evalsha` is configured to delay-then-throw-NOSCRIPT with `timeoutMs` set below the delay. Assert `fakeRedis.evalCalls === 0`. Optionally add a negative test by temporarily monkey-patching `ctx` (out of scope for a unit test; the positive assertion is sufficient). Not blocking.

### Security (STRIDE) — no findings

Oracle-2 verified Spoofing, Tampering, Repudiation, Information Disclosure, DoS, and Elevation of Privilege end-to-end. Notable confirms:

- `invalidateRedisClient`'s null-before-disconnect ordering (`redis.ts:43–47` → `:49–54`) means a thrown `disconnect()` cannot leave a poisoned singleton.
- WeakSet dedup correctly removes its entry on throw (`:29`) so retries aren't permanently blocked — verified by `rate-limit.test.ts` "allows disconnect cleanup to retry after a synchronous failure".
- Cleanup error logs (`:53`, `:74`, `:158`) are explicit and additive; the pre-existing `firstErrorLogged` gate is preserved, not weakened.
- **Cooldown on timeout-disconnect is a DoS *reduction*, not new surface** — bound is at most one new client every 5 s during an outage; warmed local fallback fails closed.
- Singleton invalidation cannot grant quota to anyone (no fail-open path; local fallback `indexOf` rollback handles dup-timestamp case from PR #242).

### Design / quality — no HIGH/MEDIUM findings

Oracle-3 verified the `disconnectRedisClientOnce` / `invalidateRedisClient` / `resetRedisClient` separation, the captured `connectPromise` local in `getReadyRedisClient`, and the singleton lifecycle under concurrent failure. The helper split is justified by the three distinct call sites (timeout recovery, lazy-connect failure, test teardown) and the WeakSet identity guard is correct.

### LOW / INFORMATIONAL (consolidated, not blocking)

1. **`disconnectRedisClientOnce` exported solely for one direct test** (oracle-2 + oracle-3). The `redis.ts` public API is wider than necessary. Either move that test to use `invalidateRedisClient` (which also exercises the disconnect path) or mark the function `@internal`.
2. **`setClientForTesting(client: Redis | null)` requires repeated `as never` casts** (oracle-3). Five call sites in tests. A typed overload `setClientForTesting<T extends Pick<Redis, "disconnect" | "evalsha" | "eval">>(client: T)` would remove the casts and document the test double's minimum surface.
3. **`timeoutMs?` is exported and parameterizable** (oracle-2). Today only used in tests, but a future production caller could pass a tiny value and turn normal Redis latency into timeout storms. Recommend `/** @internal */` JSDoc or a test-only `checkRedisRateLimitForTesting` overload.
4. **`disconnectingRedisClients` WeakSet is process-scoped** (oracle-2). Bounded by the 5 s cooldown; for a long-running process this is single-digit MB. Pre-existing pattern, not a defect.
5. **Missing regression tests for two interleavings** (oracle-3): (a) concurrent `invalidateRedisClient` + `getReadyRedisClient`, (b) `closeRedisClient` followed by `getReadyRedisClient`. Each would be a 10-line test. Optional follow-up.
6. **Author's prior `evalshaCalls` (mutable) vs `evalKeys` (readonly) drift** (carried from first review). Cosmetic; the public `FakeRedisClient` API works either way.

### Verified positive behaviors (no action needed)

- Singleton nulling + cooldown on timeout prevents handing a wedged client to subsequent callers.
- Stale failed-connect cannot evict a replacement singleton (`getReadyRedisClient` finally-block identity check).
- WeakSet dedup + retry-after-throw works under concurrent timeouts (`disconnectCalls === 2` assertion in `rate-limit.test.ts:248`).
- `getReadyRedisClient` correctly returns `null` during cooldown and rebuilds fresh after.
- New integration test exercises real Redis atomicity under concurrency: 5 allowed / 15 rejected out of 20 simultaneous requests with `maxRequests: 5`.
- Pre-existing F1 regression test (`rate-limit.test.ts:486–505`) confirms `evalCalls === 0` when NOSCRIPT arrives after timeout.
- All 10 CI checks green on `bba30c3`: lint-and-typecheck, CodeQL (3 jobs), docker, test, Sourcery, Kilo, Kody Code Review, Snyk.

### Verdict

**Approve for merge.** The F1 race is closed, H1/H2 hardening is sound and tested, and the security review surfaced no attack-surface changes. The one MEDIUM (no direct F1-closure regression test) and the LOW observations are optional follow-ups that don't gate merging.

🙏 Thank you for the careful iteration — the null-before-throw ordering in `invalidateRedisClient` and the captured-`connectPromise` discipline in `getReadyRedisClient` are exactly the kind of state-machine hardening that prevents the wedged-socket DoS class.

PR #291, opened 2026-07-25
fix(provider): broaden RETRYABLE_TRANSPORT_CODES (M1)

SweetSophia:
## What

Addresses M1 from the second-pass review of PR #289 plus Sherina's review (#5081103526) on this PR. The transient-transport-code broadening is correct, but the cert codes were misclassified as retryable when they are permanent configuration faults — no handshake retry fixes them. This PR now:

- Splits the codes into two sets by fault taxonomy:
  - **`RETRYABLE_TRANSPORT_CODES`** (15 codes) — transient transport-layer failures that can self-heal within the retry budget. Per ADR §7 these classify as `provider_network`: the worker retries and queries degrade to lexical fallback.
  - **`NON_RETRYABLE_TLS_CERT_CODES`** (10 codes, incl. `CERT_NOT_YET_VALID` for clock skew) — permanent TLS certificate validation failures. Per ADR §7 these are correctness faults that must surface through readiness/status; they emit a distinct sanitized code so operators can distinguish a cert outage from an ordinary socket blip.

- Adds `isNonRetryableTlsCertificateError()` helper. The catch block in `requestEmbedding` now emits the sanitized code `provider_tls_certificate` (non-retryable) for cert faults, before falling through to the generic `provider_request_failed` for unknown failures.

- Fixes the lex-ordering claim from the original PR: `EAGAIN` (E-A-G) now precedes `EAI_AGAIN` (E-A-I) in the Set.

## Why

ADR §7 (lines 164–170) classifies "connection timeout, temporary network failure, or provider HTTP 408/429/5xx" as legitimate fallback triggers. Expired/self-signed/untrusted/rejected certs, hostname mismatches, and missing issuers are **configuration faults** that no retry fixes. Before this PR, a cert outage burned up to `maxAttempts` (default 8) TLS handshakes over ~95–159s of exponential backoff (`hybrid-worker.mjs:144`), with the loud signal only reaching `critical` after the full retry budget — and queries silently fell back to lexical with no operator-visible distinction.

## Files

- **`scripts/hybrid-provider.mjs`**
  - `RETRYABLE_TRANSPORT_CODES` trimmed to 15 transient codes (5 errno + 10 undici).
  - New `NON_RETRYABLE_TLS_CERT_CODES` set with 10 cert codes.
  - New `isNonRetryableTlsCertificateError()` helper.
  - Catch block in `requestEmbedding` emits `provider_tls_certificate` (sanitized, non-retryable) for cert faults.
  - Lex order fixed: `EAGAIN` precedes `EAI_AGAIN`.

- **`scripts/hybrid-provider.test.mjs`**
  - Split the previous "newly-added transient transport codes" test into two:
    - `provider request classifies transient transport codes as retryable` — 15 codes assert `provider_network` + `retryable=true`.
    - `provider request classifies TLS certificate validation failures as non-retryable with a distinct sanitized code` — 10 codes assert `provider_tls_certificate` + `retryable=false`.
  - The existing `retries only recognized transport failures` (ECONNRESET) and `leaves unknown transport codes non-retryable` (ENOTACODE) tests still pass.
  - Local `node --test`: 15/15 pass.

- **`.github/workflows/ci.yml`** — added `Run hybrid-provider transport tests` step (`npm run test:hybrid-provider`) so the regression guard is enforced in CI, not just locally.

## Behavior change (breaking for cert-fault lexical fallback)

| Code class | Before | After |
|---|---|---|
| 5 errno codes (`EAGAIN`, `EHOSTDOWN`, `ENETDOWN`, `ENETRESET`, `EPIPE`) | `provider_network` retryable=true | unchanged |
| 10 cert codes | `provider_network` retryable=true (silent lexical fallback) | `provider_tls_certificate` retryable=false (surfaced as correctness error, ADR §7 honored) |
| Unknown codes | `provider_request_failed` retryable=false | unchanged |

Operators can now distinguish a cert outage from an ordinary socket blip in logs and durable state via the distinct `provider_tls_certificate` sanitized code.

## Follow-ups (deferred)

- Decouple `fallbackAllowed` from `retryable` — separate boolean for "may query fall back to lexical" vs "should worker retry". Needed if product policy wants lexical continuity during cert outages. ADR amendment required.
- Replace synthetic `TypeError` test fabrication with deterministic local HTTPS fixtures (in-test generated certs: expired, self-signed, wrong-SAN, incomplete chain).
- Add nested-cause / `AggregateError` handling so Happy-Eyeballs dual-stack connect failures classify correctly.

## Summary

This PR splits the hybrid provider's transport error classification into two distinct categories so that permanent TLS certificate failures are no longer conflated with transient socket-level blips.

### What changed

- **`scripts/hybrid-provider.mjs`** — Removed certificate-validation error codes (`CERT_HAS_EXPIRED`, `CERT_REJECTED`, `CERT_REVOKED`, `CERT_UNTRUSTED`, `DEPTH_ZERO_SELF_SIGNED_CERT`, `ERR_TLS_CERT_ALTNAME_INVALID`, `SELF_SIGNED_CERT_IN_CHAIN`, `UNABLE_TO_GET_ISSUER_CERT_LOCALLY`, `UNABLE_TO_VERIFY_LEAF_SIGNATURE`) from the retryable set, and added `CERT_NOT_YET_VALID`. Introduced a new `NON_RETRYABLE_TLS_CERT_CODES` set that captures these as permanent, correctness faults. The `requestEmbedding` flow now throws a new `HybridProviderError` with code `provider_tls_certificate` and `retryable: false` for these failures, distinct from the existing retryable `provider_network` classification.
- **`scripts/hybrid-provider.test.mjs`** — Split the previous combined classifier test into two: one verifying transient transport codes (`EAGAIN`, `ECONNRESET`, `ETIMEDOUT`, undici timeouts, etc.) are retryable `provider_network` errors, and a new one verifying TLS certificate codes are non-retryable `provider_tls_certificate` errors. The cause message was updated to reflect the new semantic (certificate vs. transient).
- **`.github/workflows/ci.yml`** — Added a dedicated CI step to run the hybrid-provider transport tests.

### Why

Per hybrid-retrieval ADR §7, transient transport-layer failures (socket resets, DNS hiccups, undici timeouts) self-heal within a retry budget and degrade gracefully. TLS certificate validation failures are correctness faults — no amount of handshake retry will fix an expired, revoked, self-signed, hostname-mismatched, or untrusted cert. Treating them as retryable caused needless retry churn and obscured operator-visible cert outages. Surfacing them through a distinct sanitized code (`provider_tls_certificate`) lets the readiness/status layer distinguish a cert outage from an ordinary socket blip in logs and durable state.

### Impact

- Transient transport failures: behavior unchanged — still retried as `provider_network`.
- TLS certificate failures: no longer retried; surfaced as `provider_tls_certificate` so they propagate to readiness checks instead of being masked by retry loops.
- CI: explicit run of the hybrid-provider test suite to guard this classification going forward.


SweetSophia:
## Consolidated review — PR #291

**Verdict: request changes — one Medium, merge-blocking design issue; the rest is Low/nit.**
Reviewed head `ff23036` (both commits). Verified locally: `node --test scripts/hybrid-provider.test.mjs` → 14/14 pass; CI green.

Heads-up on provenance: the kody and codex bot reviews were made against the **stale first commit** `0bb377f` (which contained two fabricated code names, fixed by `ff23036`). Their underlying concern nevertheless survives the follow-up — details in "Bot comment validity" below.

---

### 1. [Medium — merge-blocking] Certificate-validation codes should not be classified retryable

`scripts/hybrid-provider.mjs:19-44`

All nine TLS/cert codes added here are **real Node.js emissions** — verified against the Node 24 TLS docs, the fixed 28-code allow-list in `deps/ncrypto/ncrypto.cc` (`X509Pointer::ErrorCode`), and undici's pass-through of TLS errors as `TypeError("fetch failed", { cause })`. The `ff23036` follow-up correctly replaced the two fabricated names (`ERR_TLS_CERT_HAS_EXPIRED`, `ERR_TLS_CERT_REJECTED` — zero occurrences in the Node tree). Code-name accuracy is not the problem.

The problem is the **fault taxonomy**: retry is only rational for faults that can self-heal within the retry budget (transient faults). Expired/revoked/rejected certs, hostname mismatches, self-signed/untrusted chains, and missing issuers are **permanent configuration faults** — no handshake retry fixes them. Consequences, traced through both consumers:

- **Worker** (`scripts/hybrid-worker.mjs:144`): `terminal = !(error instanceof HybridProviderError && error.retryable) || attempt_count >= maxAttempts`. A cert fault now burns up to `maxAttempts` (default 8, `hybrid-provider.mjs:6`) TLS handshakes over ~95–159s of exponential backoff (`computeRetryDelayMs`, `hybrid-provider.mjs:279-283`), logged at **warn** level (`hybrid-worker.mjs:159-165`), before going terminal anyway. The health check only fails on `critical` (`scripts/check-hybrid-worker-health.mjs:9-10`), so the loud signal is delayed by the full retry budget — per queued job, not once per broken endpoint (retry amplification).
- **Recall** (`src/lib/memory/hybrid-retrieval-runtime.ts:321-326`): retryable → `HybridLexicalFallbackError` → `NoosphereProvider.search` catches it, warns, and serves lexical with `hybridFallbackReason: "provider_network"` (`src/lib/memory/noosphere.ts:195-273, 734-749`). Non-retryable → `HybridCorrectnessError` → escapes to the orchestrator, which returns zero results + `providerMeta.error` with HTTP 200 (`src/lib/memory/orchestrator.ts:336-368`, `src/lib/memory/api/recall.ts:211-231`). So the `retryable` boolean currently decides **two independent policies** — "should the worker retry?" and "may a query fall back to lexical?" — and this PR forces them to the same answer.
- The ADR already takes a side: only temporary network failures may fall back; configuration faults must surface (`docs/HYBRID-RETRIEVAL-ADR.md:164-170, 351-359`). This change contradicts it.

The PR body's "operationally equivalent for fallback purposes" framing is the conflation to avoid. Suggested direction:

- Move the nine cert codes into an explicit non-retryable classification (e.g. `NON_RETRYABLE_TLS_CERT_CODES`) emitting a distinct sanitized code like `provider_tls_certificate` instead of the generic `provider_request_failed` — today operators cannot distinguish a cert outage from an ordinary socket blip in logs/durable state.
- If product policy still wants lexical continuity during cert outages, model `fallbackAllowed` **independently** of `retryable` rather than overloading one boolean.
- The five errno additions (`EAGAIN`, `EHOSTDOWN`, `ENETDOWN`, `ENETRESET`, `EPIPE`) are genuine transient transport faults and belong in the retryable set — that part of the PR is correct.

### 2. [Low] Synthetic tests prove set membership, not real `fetch` emission

`scripts/hybrid-provider.test.mjs:261-296` fabricates `TypeError("fetch failed", { cause: { code } })` — exactly the one-level shape the classifier consumes (`hybrid-provider.mjs:305-308`). That is the same fabrication pattern that let two fake code names pass review in `0bb377f`. Residual risk is low now (all nine names confirmed real in Node source; four were empirically probed against badssl.com per the `ff23036` commit message), but the test can never catch a future fake code. Prefer deterministic local HTTPS fixtures (in-test generated certs: expired, self-signed, wrong-SAN, incomplete chain) for the probeable cases; keep synthetic tests for the rest.

Related facts worth recording in a comment near the set:

- `CERT_REVOKED` is real but **practically unreachable**: Node performs no CRL/OCSP check by default, so `revoked.badssl.com` handshakes successfully (nodejs/node#16338). Dead-but-harmless entry.
- If cert-family coverage is the goal (in whichever bucket), `CERT_NOT_YET_VALID` (clock skew — arguably the *most* transient cert fault) and `UNABLE_TO_GET_ISSUER_CERT` are real codes currently absent.

### 3. [Low — pre-existing] No nested-cause or `AggregateError` handling

`isRetryableTransportError` checks only `error?.cause?.code ?? error?.code` (`hybrid-provider.mjs:305-308`). Happy-Eyeballs dual-stack connect failures surface as an `AggregateError` cause (no top-level `.code`), so e.g. `ECONNREFUSED` against a `localhost` provider resolving to both `::1` and `127.0.0.1` classifies **non-retryable** — the opposite of this PR's intent. Not introduced here, but it directly caps the value of broadening the set. Follow-up candidate: walk `AggregateError.errors` and/or the cause chain.

### 4. [Low] `test:hybrid-provider` is not enforced by root CI

The script exists (`package.json:9-10`) and `npm run test` runs it first, but `.github/workflows/ci.yml:114-132` runs only memory → persistence-layer → api → wiki → rate-limit → security. This PR's regression guard (including the good `ENOTACODE` non-retryable test) only executes if someone runs it locally. One line in the test job fixes it.

### 5. [Nit] PR description is stale relative to `ff23036`

- "eight additional transient error codes" — actually **14** codes are added (5 errno + 9 TLS).
- The summary still describes "the three `ERR_TLS_CERT_*` codes" — only one remains.
- The claimed "strict alphabetical ordering" invariant is not lexicographic: `EAI_AGAIN` precedes `EAGAIN` at lines 25-26 (`'I'(0x49) > 'G'(0x47)`). Harmless for a `Set`, but the invariant claim is false.

---

### Bot comment validity

- **kody-ai inline (Bug, medium)** — *partially valid, stale specifics.* It cites `ERR_TLS_CERT_HAS_EXPIRED` / `ERR_TLS_CERT_REJECTED`, which no longer exist in this PR (removed by `ff23036`), and its suggested patch strips **all** TLS codes including the legitimately-debatable ones. The core concern (cert faults ≠ transient) is correct and matches finding 1.
- **codex P1 "Keep certificate validation failures non-retryable"** — *valid concern, overstated impact.* Confirmed: `hybrid-retrieval-runtime.ts:321-326` does degrade retryable provider errors to lexical. Overstated: a recall query makes **one** provider call — `maxAttempts` applies to worker jobs, not queries; hybrid cache hits return before any embedding call (`hybrid-retrieval.ts:203-220`); and the degradation is not literally silent (warn log + `hybridFallbackReason` metadata), though clients reading only injected text will never see the metadata. "P1" is too hot for an optional, off-by-default feature; Medium is right.
- **kody "no issues found" + sourcery "looks great" top-level reviews** — *not credible.* Both missed the taxonomy issue and the (then-present) fabricated code names.
- Note that `ff23036` already resolved the highest-severity fact either bot adjacent-touched (fake code names) — their comments predate it.

### What's good

- The `ff23036` follow-up is exemplary: empirical verification against badssl.com with per-endpoint results in the commit message, and it caught real fakes.
- The `ENOTACODE` non-retryable regression test is exactly the right guard against over-broadening.
- errno additions are correct, tested (`test.mjs:267-281` covers all five), and the unknown-code default stays fail-closed.

**Bottom line:** ship the five errno codes; reclassify the nine cert codes as non-retryable (ideally with a distinct sanitized error code); add `test:hybrid-provider` to CI; refresh the PR description. Findings 2–3 can land as follow-ups.

